PR #36, opened 2018-03-13
fixed disabled VACUUM button issue

yaozhihang:
When launching the import-export-tool and then connecting to a PostGIS database, The VACUUM button in the Database panel cannot be enabled automatically. This issue is new and does not appear in the release version 3.3.2. This pull request attempts to fix this issue.


clausnagel:
Thanks for reporting. The `setEnabled` method of `IndexOperation` is already invoked by `DatabaseOperationsPanel` and transitively from `DatabasePanel`. So invoking it inside the `handleDatabaseConnectionStateEvent` method is in fact redundant and moreover should not be required.

Turned out that the cause of the issue was the order in which the the different database operations are registered with the event bus. The `handleDatabaseConnectionStateEvent` method of `IndexOperation` must be invoked before the `setEnabled` method. This had to be fixed in `DatabasePanel`. Somehow this part of the class had been changed after version 3.3.2.

Issue #63, opened 2018-11-15
Support export of ZIP files


clausnagel:
The CityGML exporter should be able to create ZIP files containing the exported CityGML dataset and all texture files.

clausnagel:
Support for exporting CityGML content as (G)ZIP files has been developed in #66 and merged with 59c11d55543fa37218e16c4893fb31de13c8f3c3.

Issue #62, opened 2018-11-15
Support import of ZIP files


clausnagel:
The CityGML importer should be able to read ZIP files containing CityGML files and textures.

clausnagel:
Support for importing CityGML content from (G)ZIP files has been developed in #66 and merged with 59c11d55543fa37218e16c4893fb31de13c8f3c3.

PR #66, opened 2018-11-20
ZIP and GZIP support for CityGML imports and exports

clausnagel:
Solves issues #62 and #63


clausnagel:
Well, we did a lot of tests, and had to change the intially proposed implementation. But now reading/writing support for (G)ZIP is stable enough now to be merged.

PR #68, opened 2018-11-21
Global external reference type

yaozhihang:


clausnagel:
Looks good, thanks. Tested with the following XML query. The Importer/Exporter uses the `3dcitydb-schema.xml` in the background to dynamically translate the XML query to a SQL query. Worked fine.

```xml
  <typeNames>
    <typeName xmlns:core="http://www.opengis.net/citygml/2.0">core:_CityObject</typeName>
  </typeNames>
  <filter>
    <propertyIsEqualTo>
      <valueReference>core:externalReference/core:externalObject/core:name</valueReference>
      <literal>0003000a000afb11</literal>
    </propertyIsEqualTo>
  </filter>
</query>
```

Love the new XML query functionality of the exporter :-)

PR #78, opened 2019-02-08
Add support for non-GUI plugins

clausnagel:
The current Plugin API and its internal implementation suppose that each plugin provides some sort of GUI component. However, it is also useful to have plugins that are just hooked into internal processes such as CityGML imports and exports. For example, this would allow for writing a plugin that receives and thus can manipulate every CityGML feature before it is written to the output file.

This PR is to rework the Plugin API such that non-GUI plugins can be supported in principal.


clausnagel:
discussed with @yaozhihang offline

PR #85, opened 2019-04-14
added warning dialog for modules not supporting CityGML ADEs

yaozhihang:


yaozhihang:
good ideas.

just one question: Is it necessary to add the condition "at least one of the registered ADEs is supported by the Importer/Exporter" for puping up the warning dialog? I think no.

clausnagel:
No, it is not necessary. The warning could also just pop up whenever ADEs are registered in the database.

I just thought it would be consistent with the CityGML exporter. Even if ADEs are registered, the CityGML exporter would ignore them unless a corresponding ADE extension is installed.

yaozhihang:
Now ready for review

PR #86, opened 2019-04-16
Add support for sorting CityGML exports

clausnagel:
Add the possibility to sort the top-level features in a CityGML export based on one or more simple thematic attributes. It should be possible to define the sorting criteria using the XML query language, e.g. as new `<sortBy>` element. Implementing this feature requires the following:

1. Currently, the mapping of export filters to a corresponding SQL query makes use of `UNION` and `INTERSECT` to map, for instance, logical operators. This impedes sorting the entire result set. Thus, the mapping should be refactored to create a more concise SQL query.
2. The export process is multithreaded. Even if the database returns the features to be exported in an ordered sequence, this order gets lost. So, this needs to be refactored without replacing the multithreaded approach.

Note that sorting will only work on the requested top-level features. It will not work on additional objects that are part of the export even if those objects are also top-level. For instance, when requesting `<Building>` features in a sorted manner, then the resulting dataset may also contain global appearances of the buildings as additional top-level features. The sorting criteria obviously cannot be applied to these appearance elements though.


FxKu:
83 files changed only for sorting?! Wow. Just a quick question as I couldn't find it in the code: Is sorting by a geometry (e.g. envelope) also possible? In PostGIS this will align the objects on a morton curve.

clausnagel:
@FxKu some files just changed because my IntelliJ reorganized imports, but nevertheless adding support for sorting indeed required some major changes.

Currently sorting is restricted to thematic attributes. But in fact there is no reason why to exclude spatial columns. Is it enough to just add the spatial column in the ORDER BY clause?

FxKu:
Yes. But the Morton-Key sorting behavior only exists in [PostGIS 2.4+](http://blog.cleverelephant.ca/2017/09/postgis-operators.html) and afaik not in Oracle.

clausnagel:
Merging now with support for sorting based on thematic attributes. Support for spatial sorting will follow later.

Issue #89, opened 2019-04-29
Support export of CityJSON files


clausnagel:

clausnagel:
added with PR #166

PR #92, opened 2019-06-18
Add support for importing CityJSON datasets

clausnagel:
This PR relates to #88. A first version of the CityJSON import is ready for review.


yaozhihang:
@clausnagel cool work!
One thought: It seems that the CityJSONReader read the CityJSON file into main memory, which may leads to `java.lang.OutOfMemoryError` when importing large CityJSON file.  Is it possible to use chunk-wise read&Import like the CityGML-Import?

clausnagel:
Thanks @yaozhihang. Yes, you are right, each CityJSON dataset is read into main memory for processing. There are deliberate design decisions in CityJSON that hinder a similar chunk-wise parsing of input datasets as we have implemented it for GML-encoded CityGML datasets. Also see this CityJSON issue: https://github.com/tudelft3d/cityjson/issues/6. 

A good overview of pros and cons of the CityJSON approach can be found here: https://github.com/tudelft3d/cityjson/issues/20.

Being able to read a CityJSON dataset feature-by-feature would help us to keep the memory footprint of the import process low. The CityJSON approach is rather to keep the size of individual files small, and then to link several files by a master file (very much like the `tileset.json` file is used in [3D Tiles](https://docs.opengeospatial.org/cs/18-053r2/18-053r2.html)). This is of course also a valid approach.

To avoid out of memory errors while parsing a CityJSON file, we could add a simple file size check and refuse processing files that exceed a certain limit (not sure what this limit is though). Any other ideas are welcome!

For the export of CityJSON files (see #89) we will face a similar challenge. The current CityGML export simply streams every city object into the output file. For the CityJSON export, we have to make sure that the output file doesn't grow too big. One approach could be to limit the number of city objects per file. Or we could enforce a spatial tiling for every CityJSON export, with every tile being written to a separate CityJSON file. The second approach would be easier to implement based on the exisiting code. Well, again, I'm happy for ideas.

clausnagel:
Hi @hugoledoux, nice to meet you here.

1. Yes, XLink information is temporarily stored in a database (either server-side or in a local file-based database). The XLinks are resolved in a separate step after the import. This allows us to handle forward and backward XLinks in arbitrarily large files, but requires a two-phase import approach.

2. For textures (especially global appearances) it's the same. 

3. We already offer spatial tiling or a maximum feature count for CityGML exports to keep the output file sizes small. But if the user doesn't define this, we currently do not have to take special care because we then simply stream into a single (and potentially huge) file. The user must know what to do with this result and how to further consume it.

    This fallback strategy would have to be different for CityJSON. But that's fine, just takes additional efforts. The master-file approach you have in mind would be a way to handle this (if you already have material to share I would be interested). The only additional issue I see is how this would work with a WFS service. 

4. Yes, falling back to cjio would be an option. cjio could be either bundled with the Importer/Exporter, or (as a first simple solution) we could point the user to cjio in a user dialog if the CityJSON file size is too big.

Well, I did some more experiments with large CityJSON files. It seems that the [Gson](https://github.com/google/gson) library we use for reading CityJSON has a memory issue. For a 200MB CityJSON file, Gson occupies ~6GB while parsing. The memory is released once the file has been read, but the application may have already crashed beforehand. Well, not so nice, will have to take a closer look at this.

yaozhihang:
https://github.com/3dcitydb/importer-exporter/blob/cityjson-import/impexp-core/src/main/java/org/citydb/citygml/importer/reader/cityjson/CityJSONReader.java#L56
I'm not sure if its is bug. The code line above shows that the global appearances are counted for the feature counter filter. But according to the documentation (p.138), the counter filter should only be applied for top-level features. The CityGMLReader may have the same issue.

clausnagel:
@yaozhihang I did not change the behaviour compared to the `master` branch. Global appearances are counted there as well. I think this is a separate discussion and independent from this PR. Would you like to go ahead and create a new issue for this?

clausnagel:
I reworked some of the Gson code in [citygson](https://github.com/citygml4j/citygson) which helped to reduce the memory footprint by the factor ~4.5. The importer can now handle reasonably large CityJSON files. There will always be an upper limit, but also other software will face this.

yaozhihang:
@clausnagel Nice. The import of the large CityJSON file is now significantly better!

I found another small issue. When running a validation process against this CityJSON dataset [Test_LOD2_Building_CityJSON.zip](https://github.com/3dcitydb/importer-exporter/files/3330010/Test_LOD2_Building_CityJSON.zip), the console window prints out the following error message:
```
[14:31:08 ERROR] An unexpected error occurred.
[14:31:08 ERROR] java.lang.RuntimeException: Failed to invoke public org.citygml4j.cityjson.feature.AbstractCityObjectType() with no args
[14:31:08 ERROR] java.lang.InstantiationException: null
``` 
Nevertheless, the CityJSON-Import works very well.

clausnagel:
fixed in 729862fb231abf9c8dfe2cb13c811d076dfee32d

Issue #93, opened 2019-06-18
No KML export of BuildingInstallations modelled as ImplicitGeometry?


gioagu:
Dear developers,

A question to you, as I could not find it in the documentation (or I may have missed it) of the 3DCityDB (3.x).

I have a number of LoD2 buildings. Additionally, they have some (outer) BuildingInstallations which are modelled geometrically via lod2ImplicitRepresentation (using MultiSurfaces).

I can validate the CityGML input file in the 3DCityDB importer, I can import it correctly into the 3DCityDB, but I cannot export the BuildingInstallations using the 3DCityDB KML exporter.

Is is a bug or it is just not supported?

If I model each BuildingInstallation "as usual" via lod2Geometry (again, using MultiSurfaces), than it the exporter works as expected.

If needed, I can provide an exerpt of the CityGML file. Let me know!

clausnagel:
Hi @gioagu, well, it should work but it seems like we never tested this scenario. Would be great if you can provide us some test data so that we can fix this issue. Does the CityGML export work correctly?

gioagu:
Here you go.
This is exported from the 3DCityDB. I'd say, yes, the exporter works.
There are for each building 3 BuildingInstallations. They look like spheres on top of the buildings.
The lowest one is modelled as lod2Geometry, the other 2 as lod2ImplicitGeometries (Solid and MultiSurface, respectively).
In FZK you can visualise them easily.

The KMZ exporter only exports the first one.

The zipped CityGML file is attached: [Test_mini_out.zip](https://github.com/3dcitydb/importer-exporter/files/3306770/Test_mini_out.zip)

clausnagel:
Thanks @gioagu. I checked the current implementation. Implicit geometries are only considered by the KML/COLLADA/glTF exporter if they are modelled for top-level features (such as `frn:CityFurniture` or `gen:GenericCityObject`) but not for nested sub-features such as `bldg:BuildingInstallation`. 

So, the KML/COLLADA/glTF exporter needs to be enhanced to support your test dataset correctly. I am going to work on it, but it may take a bit more time. I will post updates here.

clausnagel:
Fixed in `master` branch (1177b6657fc963cbb51c3fd1f5aadb32f5189757). Below the exported model is shown in Google Earth.

![ge](https://user-images.githubusercontent.com/5517439/60253683-da9d4100-98cc-11e9-855a-54a62821631f.png)

clausnagel:
[...]

PR #94, opened 2019-06-20
Add support for implicit geometries of nested sub-features in KML/COLLADA/glTF exports

clausnagel:
The current implementation of the KML/COLLADA/glTF exporter supports implicit geometries only for top-level features such as `frn:CityFurniture` or `gen:GenericCityObject` but not for nested sub-features such as `bldg:BuildingInstallation` (also see #93).

This PR enhances the KML/COLLADA/glTF exporter to also export implicit geometries of nested sub-features.

One simple test dataset: [Test_mini_out.zip](https://github.com/3dcitydb/importer-exporter/files/3311150/Test_mini_out.zip)


clausnagel:
Good point, Zhihang. I agree that we should postpone general support for point and curve implicit geometries to a future PR. I think it’s neither fully supported for implicit geometries of top-level features currently. 

We should however make sure to put this into the documentation for the next release.

Son-HNguyen:
The PR looks good to me as well.

PR #96, opened 2019-07-01
Enable BBOX calculation using CLI and add this to KMLExport

Son-HNguyen:


impexp-client/src/main/java/org/citydb/cli/ImpExpCli.java
```diff
@@ -188,6 +194,12 @@ public boolean doKmlExport(String kmlExportFile) throws ImpExpException {
 
 		log.info("Initializing database export...");
 
+		// automatically calculate bounding box --> will overwrite bounding box defined in the config file
+		BoundingBox configBBOX = config.getProject().getKmlExporter().getQuery().getBboxFilter().getExtent();
```
clausnagel: Why is the bounding box given in the config file always overriden with a calculated one for KML exports? I don't think that this is a good idea. Maybe there could be an additional command line parameter to control the behaviour of the KML export if required.

impexp-client/src/main/java/org/citydb/cli/ImpExpCli.java
```diff
@@ -299,4 +311,118 @@ private void initDBPool() throws ImpExpException {
...
+		BoundingBox bbox = null;
+		DatabaseConnectionPool dbConnectionPool = DatabaseConnectionPool.getInstance();
+		try {
+			log.info("Calculating bounding box...");
+			if (dbConnectionPool.getActiveDatabaseAdapter().hasVersioningSupport())
+				return null;
+
+			try {
+				FeatureType featureType = schemaMapping.getFeatureType(config.getProject().getDatabase().getOperation().getBoundingBoxTypeName());
+				//FeatureType featureType = config.getProject().getKmlExporter().getQuery().getFeatureTypeFilter().
```
yaozhihang: Maybe the feature type filter shall be taken into account here for calculating the bounding box when running KML-Export?

impexp-client/src/main/java/org/citydb/cli/ImpExpCli.java
yaozhihang: Right. The bounding box in the config should not always be overriden by a calculated one. 
How about adding an condition check to say: if the bounding box has not defined in the config file, then the config should be automatically filled with a calculated one?

clausnagel:
Good input, @yaozhihang. In fact, the `calcBoundingBox` method of the database util adapter is already available from the `impexp-core` module. I guess not much more is needed here (note that the coordinate transformation in `doCalcBBOX` is pointless due to 1) `targetSrs = bboxSrs` and 2) `targetSrs.getSrid() != dbSrs.getSrid()`).

impexp-client/src/main/java/org/citydb/cli/ImpExpCli.java
Son-HNguyen: @clausnagel Added a CLI flag `-calcBBOX` for auto bounding box calculation (see https://github.com/3dcitydb/importer-exporter/commit/dd90f7c26f41f6078ef36147f35fccc2f8aa4a72)

Son-HNguyen: @yaozhihang If a bounding box is not defined in the config file, it shall automatically be calculated from the database (see https://github.com/3dcitydb/importer-exporter/commit/8372709505f045f6d9930b2c16938c245aa1f834)

Son-HNguyen: For example the commands:
+ `java -jar lib/impexp-client-4.2.2.jar -shell -kmlExport "export/export.kml" -calcBBOX` 
shall automatically calculate the bounding box regardless of whether a bounding box already exists in the config file or not
+ `java -jar lib/impexp-client-4.2.2.jar -shell -kmlExport "export/export.kml"` shall not override the bounding box in the config file, however if such bounding box does not exist, it shall be automatically calculated from the database

Son-HNguyen:
Thanks for the input @yaozhihang @clausnagel 

I've adjusted the bbox function a little bit (see https://github.com/3dcitydb/importer-exporter/commit/330c51e793484c8cb9131c795233754bcc3f30e6).

I also moved the function `getObjectClassIds()` from class `BoundingBoxOperation` to `org.citydb.util.Util` (see https://github.com/3dcitydb/importer-exporter/commit/19da04baf626449e43cce6de9983085aa90d525c).

The function `calcBoundingBox()` in class `BoundingBoxOperation` and the function `doCalcBBOX()` in class `ImpExpCli` are however not exactly the same, since the former interacts with the GUI while the latter does not (because it is for CLI). Therefore, I didn't merge both of them into a `public static` util method.

clausnagel:
This PR is open for quite some time now. The CLI will be reworked substantially soon, so this PR should serve as input for this upcoming development.

On the other hand, I think this PR addresses a more general enhancement than just adding another parameter to the CLI. A KML/COLLADA/glTF export is currently only possible if you either provide a list of gml:ids or a BBOX. I think both should be optional and if omitted, the exporter should calculate the required BBOX automatically in the background. Also users of the UI would then benefit from this enhancement. 

So, my recommendation is:
1. to close this PR without merging
2. to open an issue for automatically calculating the BBOX for KML/COLLADA/glTF exports if the user has not provided one
3. to make this option available in the new CLI

I think (2) will need some UI rearrangements. And automatically calculating a BBOX should rather be the default option, so that an explicit command-line parameter (3) might even be unnecessary. Thus, a solution for (2) would even be independent from the work on the new CLI.

Thoughts? Objections?

Son-HNguyen:
Of course, if the CLI will be reworked soon, there's no need to merge this branch to the current `master`.

I was thinking the same about the automatic BBOX calculation for KML/COLLADA/glTF exports while implementing this branch some time ago. Will have a look at it ASAP.

yaozhihang:
@clausnagel I like your proposal. 
One thought: should the bbox be calculated based on all contents in database, or those features that satisfy the user-defined filters e.g. feature type filter?

clausnagel:
@yaozhihang, well, I think if you have a filters defined, the BBox should only be calculated for the matching features. Only if no filter is specified, the BBox should be calculated based on all content in the database. Would you agree?

clausnagel:
Closing now based on unanimous consent. The disucssion on how to automatically calculate the BBOX for KML/COLLADA/glTF exports will take place outside this PR.

Issue #95, opened 2019-06-26
Counter filter failed when importing CityGML data having global appearances


yaozhihang:
The Import process failed to import e.g. the first 5 building objects of this test dataset [LOD2_Buildings_v200.zip](https://github.com/3dcitydb/importer-exporter/files/3330608/LOD2_Buildings_v200.zip) using the counter filter .

It seems that the Importer counts the global appearances for feature counter filter. This is not correct because according to the documentation (p.138), the counter filter should only be applied for top-level features.

PR #98, opened 2019-08-01
fixed import of CityGML file with global appearances and featureCounterFilter

yaozhihang:
This PR should fix the issue https://github.com/3dcitydb/importer-exporter/issues/95


clausnagel:
In general I do support this PR. However, with the current implmentation there is no more UI feedback when importing global appearances. Since a dataset can contain a lot of global appearances (even more than city objects), there should be some UI feedback.

yaozhihang:
You are right. I'll add some UI feedbacks in the progress dialog.

PR #116, opened 2020-03-20
use online documentation

yaozhihang:


clausnagel:
The docu does not open when starting the Importer/Exporter from my IDE, because the manifest with the docUrl parameter is only created when building the tool. Shouldn't there be at least a log message or a default behaviour (like opening the doc at https://3dcitydb-docs.readthedocs.io/en/latest/index.html?

yaozhihang:
With the commit (1a020a2) you are now able to open the docUrl, when the ImpExp is startet via IDE

PR #123, opened 2020-07-19
Provide CLI script

clausnagel:
The recommended way to use the Importer/Exporter in CLI mode is to invoke the `impexp-client.jar` using the `java` tool. From the [doc](https://3dcitydb-docs.readthedocs.io/en/latest/impexp/cli.html#):

    $ java -jar lib/impexp-client-{version}.jar [-options]

This is neither intuitive nor handy and it also faces disadvantages. For example, you cannot easily put this command on your `PATH` and the `{version}` part of the command must be adapted with every new release. In theory, you could use the provided start script `3DCityDB-Importer-Exporter(.bat)` to run the Importer/Exporter in CLI mode. But at least for Windows, the start script uses `"start /min cmd /c"` to execute the tool. This is OK and makes sense for launching the graphical user interface but suppresses all output when invoking the script from a shell.

This PR proposes to have an explicit CLI script `impexp(.bat)` in addition to the start script. The CLI script is located in the `bin` subfolder. This makes using the Importer/Exporter in CLI mode really simple. For example, the following command shows the help message and all available options:

    $ impexp -help

This way, the CLI script can be put in the `PATH` and is independent of the release version. And the start script can be re-implemented such that it just invokes the CLI script with the option to launch the GUI.


clausnagel:
Discussed offline with @yaozhihang. Merging.

PR #126, opened 2020-08-11
CityGML export improvement  

clausnagel:
CityGML export performance is fine if the Importer/Exporter and the database server are close to each other and network latency is low. Performance is dropping though if latency increases, for instance, if  the database server runs in the cloud and the Importer/Exporter on your local machine. This PR rewrites parts of the exporter to use less queries to fetch the data, which can help to reduce the export time in case the Importer/Exporter is not running on the same machine or in the same local network as the database server.

~~This PR does not change the exporter API and, thus, ADE extensions implemented against the API should still work.~~


clausnagel:
Commit 3be92def6b98b58363166616cf8e5e588f57c27e breaks the ADE exporter API and thus is not backwards-compatible

clausnagel:
Ok, I think this PR is ready for review now. I had to work on smaller changes after my initial tests. For instance, I have changed the dynamic batch size to a fixed batch size due to performance issues under Oracle (e74f52f). This is a pity, but now exports from Oracle also benefit from the changes in this PR. The batch size can be set in the config file.

My tests suggest that there is also a certain performance gain in case the Importer/Exporter and the database are close to each other and not only in case the network latency is high.

yaozhihang:
The PR looks very good. There two remarks from my first review:
1. it would be good to add UI options for setting the parameters of `exportBatching`
2. There might be a bug while resolving xLinks of `surfaceDataMember` during CityGML export. The attached CityGML file contains one building with 3 Appearances. The Import works fine. However, the two Appearances containing xilnks are missing in the exported CityGML file 
[sample1.zip](https://github.com/3dcitydb/importer-exporter/files/5214997/sample1.zip)

clausnagel:
re 1: The export batch sizes can be configured in the config XML file. I support your proposal, but suggest to decouple this GUI-related enhancement from this PR and to rather create a separate issue for it once this PR has been merged.

re 2: Thanks for your careful testing! Fixed in 8a696b4

yaozhihang:
re 1: Ok, agree. 

re 2: Thanks, good Fix!

Issue #28, opened 2017-05-11
Unsupported geometry type MULTI_GEOMETRY, Data is int CityGML format LOD3 level


SurajitChowbey:
Unable to import CityGML format data which is in LOD3 level.
I have uploaded the error log data with this issue.
[error.txt](https://github.com/3dcitydb/importer-exporter/files/992213/error.txt)

FxKu:
This is a known limitation of the Importer/Exporter. Usually, you can fix it yourself by changing all MultiGeometry properties into MultiSurface ones. See also #27.

SurajitChowbey:
Thanks, now I am getting the error but, still the Multi data is not getting displayed.
Can you please tell me where can I get this kind of information that Importer/Exporter support this not this.
It will really appreciate your help.

FxKu:
Like in #27 it helps if you can provide sample data . Objects modelled with `gml:MultiSurface` should be displayed. Unsupported features are usually documented in the manual. Not in this case afaik, so we should add this info.

SurajitChowbey:
Okay I have upladed the sample data.
link:
https://drive.google.com/open?id=0ByPmpvXvo7O2TnRPcHRnUGV5c3M

FxKu:
so I removed all the multigeometry objects and was able to see some pipes in Google Earth
![pipes](https://cloud.githubusercontent.com/assets/5940260/25956752/dad33270-366c-11e7-8574-c0bd8f251422.png)

SurajitChowbey:
Yes, it should look like that.
I'm new in this field so, I don't have good knowledge in it.
Can you please guide me what changes I need to do in the GML file and also, please send me the updated gml file, so I can understand it better.

My mail-id: **surajit.kr.chowbey@gmail.com**

Thanks for your time.

FxKu:
You are working with FME, right? Maybe you can change the CityGML write process, that it does not write these MultiGeometry properties. I simply removed the corresponding tags (incl. geometry member) from the XML file within a text editor. Got some XLink errors during the import, but the export looks ok. Try it out. Have to catch my plane, now :)

SurajitChowbey:
Thanks for your time.
I have tried a lot and still doing it, but as I new in this domain, still not getting the right result from the CityGML file.

It will be very helpful for me, if you please share the modified CityGML file.

FxKu:
Alright. I've send the file to your address.

SurajitChowbey:
Thanks, it's working.
[...]

PR #130, opened 2020-08-24
Resolve the limitation to gml:CompositeSurface as exterior shell of solid geometries

clausnagel:
The CityGML import only supports `gml:CompositeSurface` as exterior shell of a `gml:Solid`. However, according to the GML 3.1.1 XML Schema, any surface is allowed to describe the shell of a solid. This PR proposes to adapt both the CityGML import and export to resolve this limitation.

Solves https://github.com/3dcitydb/3dcitydb/issues/52

### DIFF
--- a/impexp-core/src/main/java/org/citydb/citygml/exporter/database/content/DBSurfaceGeometry.java
+++ b/impexp-core/src/main/java/org/citydb/citygml/exporter/database/content/DBSurfaceGeometry.java
@@ -98,7 +98,7 @@ public class DBSurfaceGeometry implements DBExporter {
 			if (exporter.getDatabaseAdapter().getSQLAdapter().requiresPseudoTableInSelect()) select.setPseudoTable(exporter.getDatabaseAdapter().getSQLAdapter().getPseudoTableName());
 			select.addSelection(ComparisonFactory.exists(new Select().addProjection(new ConstantColumn(1).withFromTable(table))
 					.addSelection(ComparisonFactory.equalTo(table.getColumn("surface_geometry_id"), new PlaceHolder<>()))));
-			
+
 			psImport = cacheTable.getConnection().prepareStatement("insert into " + cacheTable.getTableName() + " " + select.toString());
 		}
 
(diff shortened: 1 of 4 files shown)


yaozhihang:
Check this test data. The import process throws error messages in the console. 
[testGml.zip](https://github.com/3dcitydb/importer-exporter/files/5142875/testGml.zip)

clausnagel:
Thanks, @yaozhihang. Seems like the import of `gml:GeometricComplex` was already broken before this PR. In general, both `gml:MultiGeometry` and `gml:GeometricComplex` are not supported by the table `SURFACE_GEOMETRY` because they might contain point-like and curve-like geometries that cannot be stored in this table. Therefore, there is no way to create a root entry for them. 

The importer already tries to map a `gml:MultiGeometry` to a `gml:MultiSurface` to be able to import it (see #28). I have now extended this behaviour to all unsupported geometries (see e146c51). This way the `gml:GeometricComplex` in your dataset can be imported. But note that this mapping is not lossless in general (both geometries and appearances might get lost). So it's better to not use `gml:GeometricComplex` or  `gml:MultiGeometry` at all.

yaozhihang:
@clausnagel thanks for the quick fix.  I made some changes the test dataset so that the  `gml:GeometricComplex` contains now a `CompositeSurface`. The CityGML import throws no errors, but it seems that redudant records are inserted in the SURFACE_GEOMETRY table. Here is the test data. 
[testGml2.zip](https://github.com/3dcitydb/importer-exporter/files/5144204/testGml2.zip)

clausnagel:
Thanks for your thorough testing. This bug also exists outside this PR already, but I fixed it here. It's good to have more tests to improve the importer. @yaozhihang, maybe we should open dedicated issues and PRs in case there are more bugs that are not directly related to this PR?!

yaozhihang:
@clausnagel  you are right. Each PR should exclusively focus on its related issues. This PR now looks very good and works like a charm throughout my tests.

PR #131, opened 2020-08-31
move kmlExport ADE APIs to impexp core module

yaozhihang:
This PR moves all KmlExport-related APIs for ADEs into the impexp-core module. Thirdparty applications now just need the dependency of the core module to load ADE extension modules, when they implement the `ADEKmlExportExtension` interface.


clausnagel:
Thanks for proposing this change, which I fully support. The only thing I'm not sure about is moving `Queries.java` into the `org.citydb.ade.kmlExporter` package. This class contains all the database queries required for the KML/COLLADA/glTF export and thus is a core component of the `impexp-kml-collada-module`. I therefore think the previous package `
org.citydb.modules.kml.database` is more appropriate for this class.

I understand that we run into circular dependencies if we just keep `Queries.java` in its old place but move all other classes into the `impexp-core` module. Does an ADE implemention really need access to all the queries in `Queries.java`, or could we just split the class and move some general helper queries such as `getSurfaceGeometries` or `getImplicitGeometryNullColumns` into the core module?

yaozhihang:
Good point. You are right, the `Queries.java` should be kept in the previous package `org.citydb.modules.kml.database`.
In fact, a number of methods in `Queries.java` are required by ADE implementation. We can abstract these methods into a dedicated helper interface which can be placed in the core module. see this proposal c6f918118bcea9fa7265f0185076fc6ab7319516

clausnagel (approved):
Looks fine

yaozhihang:
Approved by @clausnagel  offline. Merge it.

PR #133, opened 2020-09-08
Enhance LoD export filter

clausnagel:
This PR proposes to enhance the LoD export filter with the option to only export the minimum or maximum LoD from the list of selected LoDs. For this purpose, two new filter modes `minimum` and `maximum` are added to the filter configuration.

The following snippet shows an example of how to use the new LoD filter modes:
```xml
  <typeNames>
    <typeName xmlns:bldg="http://www.opengis.net/citygml/building/2.0">bldg:Building</typeName>
  </typeNames>
  <lods mode="minimum" searchMode="depth" searchDepth="1">
    <lod>1</lod>
    <lod>2</lod>
    <lod>3</lod>
  </lods>
</query>
```
In this example, the exporter will query all `Building` objects having a spatial representation in LoD 1, 2 or 3. If a building has a representation in more than one of these LoDs, only the `minimum` LoD is kept and exported.

**Note that PR #126 is a mandatory prerequisite for this PR**

### DIFF
--- a/impexp-config/build.gradle
+++ b/impexp-config/build.gradle
@@ -3,7 +3,7 @@ plugins {
 }
 
 dependencies {
-    api 'org.citygml4j:citygml4j:2.11.0'
+    api 'org.citygml4j:citygml4j:2.11.2-b20200917-1733'
 }
 
 bintray {
--- a/impexp-config/src/main/java/org/citydb/config/project/query/filter/lod/LodFilterMode.java
+++ b/impexp-config/src/main/java/org/citydb/config/project/query/filter/lod/LodFilterMode.java
@@ -37,7 +37,11 @@ public enum LodFilterMode {
 	@XmlEnumValue("or")
 	OR("Or"),
 	@XmlEnumValue("and")
-	AND("And");
+	AND("And"),
+    @XmlEnumValue("minimum")
+    MINIMUM("Minimum LoD"),
+    @XmlEnumValue("maximum")
+    MAXIMUM("Maximum LoD");
 	
 	private final String value;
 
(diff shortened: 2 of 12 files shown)


yaozhihang:
Great enhancement. Only found one small bug in an edge-case test.

How to reproduce it:
1. import the attached CityGML file (one CityGML Building object with LOD 0, 2, and 3 representations) into citydb [lod0_2_3_xLink.zip](https://github.com/3dcitydb/importer-exporter/files/5219457/lod0_2_3_xLink.zip)

2. run CityGML export with the following LOD filter configuration 
image

3. The exported CityGML file should contain the explict LOD3 surface geometry, instead of an XLINK

clausnagel:
Ok, I see. Cross-LoD-Xlinks are in fact evil. Will take a look at it, but adding support for such cases makes things definitely more complex.

clausnagel:
Fixed the cross-LoD-XLink issue with 2120970.

The implementation now detects that a spatial property in the target LoD uses an XLink reference to a geometry from another LoD ("cross-LoD-XLink"), which is going to be deleted due to the filter settings. Well, unfortunately, it is not sufficient to just replace the XLink with the referenced geometry.

For example, assume that we have two cross-LoD-XLinks to the same geometry, one from the `bldg:lod2Solid` property of a building and another from the `bldg:lod2MultiSurface` property of a `bldg:WallSurface` of this building. According to CityGML conformance requirements, we are only allowed to put the geometry into the `bldg:lod2MultiSurface` of the `bldg:WallSurface` because the `bldg:lod2Solid` property must reference the geometries of the boundary surfaces.

So, the implementation should now also adhere to these requirements.

yaozhihang:
Brilliant solution! This PR now works well in all my tests.

PR #136, opened 2020-09-17
new static MAX_BATCH_SIZE variable in ExportBatching.java

yaozhihang:
With this PR, both `ExportBatching` and `ImportBatching` work in the same fashion for handling boundary case.


clausnagel:
I think the `getBlobBatchSize()` method in `DBXlinkExporterManager` should also be adapted correspondingly?

yaozhihang:
It should be already adpated correspondingly...

clausnagel:
Ha, of course. Guess I shouldn't review code on my phone while cooking for the kids :-)

clausnagel (approved):

Issue #135, opened 2020-09-16
Add UI options for setting the export batch sizes


clausnagel:
See discussion in PR #126

clausnagel:
Implemeted with PR #137. Kudos to @yaozhihang

PR #137, opened 2020-09-18
added UI preferences settings for batch export

yaozhihang:
This PR solves the issue here https://github.com/3dcitydb/importer-exporter/issues/135


clausnagel (approved):

clausnagel:
Thanks @yaozhihang, looks fine. One observation: you can enter 0 (which is an invalid value) and the value is shown even if you switch to another preferences page and back. Only if you restart the tool, the 0 is replaced with the default value. This behaviour is, however, consistent with the similar dialog for the importer. So, I am ok with merging. Maybe we can address the behaviour of the input fields afterwards?!

Issue #139, opened 2020-09-18
Seeking input on the redesign of the command-line interface


clausnagel:
The current CLI just offers command-line parameters for the operation to be executed, for instance, `-import`, `-export` or 
`-kmlExport`. All further settings such as database connection details or export filters must be provided in an external config file.

We want to redesign the CLI. The main goal is to provide more command-line parameters to be able to trigger default operations without the need for an additional config file. This should help to simplify the use of the CLI. Another goal is to allow plugins to programmatically extend the CLI with new operations.

Some initial thoughts:
- each operation is mapped to a separate command with its own parameters
- some global parameters are available for all commands, such as database connection details or logging settings
- a plugin can register its own additional command

For example, invoking a CityGML export using a bbox filter could look like this:

    > impexp export --db-host=localhost --db-name=citydb --db-user=user --bbox=1,2,3,4 my_citygml_file.gml

Well, it's obvious that we cannot map every option from the Importer/Exporter preferences to a command-line parameter. So, it must still be possible to provide a config file for these settings. But it should be optional and if omitted, default values should be chosen instead.

Which parameters would you like to see for the CLI? For example, is a `--bbox` parameter for a CityGML export like in the above example sufficient or do you also need more parameters to specify the bbox mode (overlap, inside), the CRS used for the bbox, or even a tiling based on the provided bbox?

Son-HNguyen:
Thanks for the init Claus. The GUI offers an intuitive way to interact with the software but the CLI is definitely a big plus and I'm all for it. Personally I use CLI when I'd like to do something quickly or when I want to automate something, so to be able to put some settings in a config file will help a lot (and it makes the commands shorter too). As for the CLI parameters, I'm thinking whether we could use the same options based on the XML config file stored for the client as a starting point (not necessarily all of the options there) and extend it along the way if needed? What do you think? Is there an online place where we could list the parameters together and edit them later if changes are needed?

yaozhihang:
@Son-HNguyen Do you mean to introduce a _simplified_ version (a subset) of the current XML config for CLI?

clausnagel:
Well, I would definitely avoid having two different XML configs. My idea is that you can provide a config file through the `--config` option like in the current version of the CLI. The XML structure must follow the default config file `project.xml`. You can omit any unnecessary XML elements and keep it as small as possible (missing elements will then be populated with default values).

[...]

PR #147, opened 2020-11-13
New command-line interface

clausnagel:
This PR proposes a completey redesigned CLI for the Importer/Exporter.

Main goals for the redesign are:
1. The current CLI commands require a full XML config file to be passed as parameter and do not work without such a file. With the new CLI, each command provides a set of parameters that allow to run the operation without a config file. The parameters do not cover each and every option of the XML config, but are meant to support the most typical scenarios. Thus, a config file can still be provided if required. If both parameters and a config file are passed to the command, the values of the parameters take precedence over the corresponding values in the config file.
2. Plugins can programmatically add their own commands to the CLI. Thus, they are not restricted to the GUI anymore but can also be invoked from the command line. You can even write CLI-only plugins.

This PR is linked to the disucssion in this issue: https://github.com/3dcitydb/importer-exporter/issues/139


impexp-client/src/main/java/org/citydb/ImpExpLauncher.java
```diff
@@ -0,0 +1,76 @@
...
+    public ImpExpLauncher withArgs(String[] args) {
+        if (args != null) {
+            this.args = args;
+        }
+
+        return this;
+    }
+
+    public void start() {
+        int exitCode = new ImpExpCli()
```
yaozhihang, line 68: should the `new ImpExpCli()` be replaced by `impExpCli`?

impexp-core/src/main/java/org/citydb/util/CoreConstants.java
```diff
@@ -47,4 +47,6 @@
     public static final String EXPORT_STUB = "exportStub";
     public static final String EXPORT_AS_ADDITIONAL_OBJECT = "additionalObject";
     public static final String UNIQUE_TEXTURE_FILENAME_PREFIX = "tex_";
+
+    public static boolean IS_GUI_MODE = false;
```
yaozhihang, line 51: It seems that all variables in this class should be _constant_. Maybe the `ObjectRegistry` is a better place for hoding this global mode?

clausnagel: Yes, I agree, it's not a perfect palce :-) I think we can get rid of this flag completely, so that's why I moved it here temporarily.

impexp-client/src/main/java/org/citydb/ImpExpLauncher.java
clausnagel: fixed with d744616

impexp-client/src/main/java/org/citydb/cli/DeleteCommand.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        return 0;
+    }
+
+    @Override
+    public void preprocess(CommandLine commandLine) {
+        if (queryOption != null && deleteListOption != null) {
+            throw new CommandLine.ParameterException(commandLine,
+                    "Error: Query options and delete file options are mutually exclusive (specify only one)");
```
yaozhihang, line 119: The querying settings in the config file and the delete file options are also mutually exclusive. When the option `--config` is specified, maybe the CLI should prints some warning/hints.

Question: why the query options and the delete file are mutually exclusive, cannot be combined?

impexp-client/src/main/java/org/citydb/cli/options/deleter/QueryOption.java
```diff
@@ -0,0 +1,175 @@
...
+
+public class QueryOption implements CliOption {
+    @CommandLine.ArgGroup(exclusive = false)
+    private TypeNamesOption typeNamesOption;
+
+    @CommandLine.ArgGroup
+    private ResourceIdOption resourceIdOption;
+
+    @CommandLine.ArgGroup
+    private DatabaseIdOption databaseIdOption;
```
yaozhihang, line 59: maybe also add the `databaseIdOption` to the CityGML export command?

yaozhihang:
We have to consider whether the `featureVersionFilter` option should be added to the `export` and `delete` CLI command.

impexp-plugin-api/src/main/java/org/citydb/plugin/cli/CliOptionBuilder.java
```diff
@@ -0,0 +1,309 @@
...
+                    if (module == null || !module.hasFeature(parts[1])) {
+                        throw new CommandLine.ParameterException(commandLine, "Unknown type name: " + typeName);
+                    }
+
+                    featureTypeFilter.addTypeName(new QName(namespace, parts[1]));
+                } else if (parts.length == 1) {
+                    Module module = Stream.concat(CityGMLVersion.v2_0_0.getCityGMLModules().stream(),
+                            Modules.getADEModules().stream())
+                            .filter(m -> m.hasFeature(parts[0]))
+                            .findFirst()
```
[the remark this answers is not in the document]
clausnagel: My initial rationale was to avoid finding the same feature type name in two ADE modules where the first one is associated with CityGML 2.0 and the second one with CityGML 1.0. Mixing feature types from different CityGML versions in one feature type filter is not supported, and this was meant as a fail-fast check for such situations.

But you are right, the check is too restrictive because it excludes valid situations. Fixed with 909b4bf4a333e593104c5e1b2b65e405d7d1f04a

impexp-client/src/main/java/org/citydb/cli/DeleteCommand.java
clausnagel: 1. Yes, technically we could combine a delete list with query options. But I think the typical use case is to use either a delete list or a query expression. I would prefer to keep them separate and mutually exclusive. 

2. The CLI options always take precedence over corresponding settings in a provided config file. This is true for all options (not just query options) and all commands. This is a correct and intended behaviour, so I don't think we should record warnings in the log. But I agree that this must be explained, and I think the documentation is the right place to do so.

impexp-core/src/main/java/org/citydb/citygml/deleter/util/DeleteListParser.java
```diff
@@ -0,0 +1,214 @@
...
+                    throw new NoSuchElementException();
+                }
+
+                String[] columns = line.split(delimiter);
+                if (columns.length < index) {
+                    throw new IOException("Found " + columns.length + " columns in delete list but expected " +
+                            "the id value in column " + index + ".");
+                }
+
+                id = columns[index - 1];
```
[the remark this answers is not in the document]
clausnagel: added with 8453bda15c52857e5104b3c7345403789bcd8334

clausnagel:
> We have to consider whether the `featureVersionFilter` option should be added to the `export` and `delete` CLI command.

I am not sure what you mean with `featureVersionFilter`. The CityGML version?

impexp-client/src/main/java/org/citydb/cli/options/deleter/QueryOption.java
clausnagel: Added with a6fab8e1663128fbcc905c9c4f9a454b8f370bdd

It should maybe then also be available for a KML/COLLADA/glTF export? But since the corresponding config settings currently do not support this, I would prefer a separate issue on this if we would like to have this option.

yaozhihang:
I mean this config option: 
https://github.com/3dcitydb/importer-exporter/blob/cli-rework/impexp-config/src/main/java/org/citydb/config/project/query/simple/SimpleFeatureVersionFilter.java

impexp-client/src/main/java/org/citydb/cli/DeleteCommand.java
yaozhihang: 1. There is an important use case of using delete list against _versioned_ database. In this case, the combination of  `featureVersionFilter` and delete list would be required. 

2. good

impexp-core/src/main/java/org/citydb/util/CoreConstants.java
yaozhihang: Good, you want to get rid of this flag in this branch or after merging into master?

impexp-client/src/main/java/org/citydb/cli/DeleteCommand.java
clausnagel: But in your example, wouldn't an additional `--exclude-terminated` flag be sufficient?

impexp-client/src/main/java/org/citydb/cli/options/deleter/QueryOption.java
yaozhihang: agree

impexp-client/src/main/java/org/citydb/cli/DeleteCommand.java
yaozhihang: `--exclude-terminated` might not be sufficient enough for the case that a user want to physically delete some very old feature versions based on certain timestamps?

clausnagel: Hm, but if your use case is really to identify the features to be deleted in such a precise way, then you can always go with an XML-based query expression or even directly use a SQL filter. For me, the use case for a delete list is to delete/terminate the current feature version only, not old versions. 

But yes, this understanding is in fact debatable. Maybe we should ask for input from others?

impexp-core/src/main/java/org/citydb/util/CoreConstants.java
clausnagel: This flag is only used to decide whether or not to calculate the total number of features affected by the export/delete operation. This calculation uses an `SQL COUNT()` based on the provided filter and, thus, can be quite slow on large databases. In GUI mode, we always apply this calculation because it allows us to render a nice progress bar. In CLI mode we suppress it in favor of a faster export/delete (especially desirable if you embed the CLI in larger processes and workflows).

Getting rid of this flag requires to implement a new option (both in the GUI and the CLI) for the user to decide whether the number of affected features should be calculated. And if it should be calculated, then my original intention for the CLI mode was to also show a nice progress bar (e.g., https://github.com/ctongfei/progressbar). But I skipped this idea for this PR because it would have taken more time to implement.

So, getting rid of the flag requires more work. I am happy to keep the flag where it is for now. What do you think?

yaozhihang: agree, this can be done in a separate pull request.

impexp-client/src/main/java/org/citydb/cli/DeleteCommand.java
yaozhihang: yes, some inputs from others are helpful, especially from the view of users, not developers :-)
I think we could leave it for now. Your currently implmentation already covers most use cases. We can change the behaviour in future when we really need such precise delete in practical use cases.

impexp-client/src/main/java/org/citydb/cli/options/vis/DisplayOption.java
```diff
@@ -0,0 +1,138 @@
...
+                visibleTo = visibleFrom;
+            }
+        }
+
+        candidates.values().forEach(displayForm -> {
+            if (!displayForms.contains(displayForm)) {
+                displayForms.add(displayForm);
+            }
+        });
+    }
```
Son-HNguyen, line 95: The current implementation requires for example the following syntax:

```
--display-mode=collada,geometry --visible-from=extruded=120,geometry=120,collada=120
```

The modes given in `--display-mode` are prioritized over those given in `--visible-from`, which means in this case `extruded=120` will be ignored. Would it be better/faster to exclude `--display-form` and use the modes given in `--visible-from` instead, so that users do not need to enter the modes twice? In this case perhaps `--visible-from` would be renamed to something like `mode-visibility`, etc.

Otherwise I'd suggest adding the following line (or similar) to the description of `--visble-from` in line 51:

```
To define modes, use --display-mode.
```

so that it is clear to the users that `--visible-from` does not automatically add modes to the display forms, only those given in `--display-mode` are accepted.

BWibo:
I am currently doing some testing and so far everything works just fine! I hope I'll be able to finish this by the end of the week.

I have one question regarding the build process of the ImporterExporter:
When creating Docker images it is important to keep those images as small as possible. I just pushed a new ``Dockerfile`` for the new CLI using multi stage build for size reduction. I ended up at ~400MB which seems OK for an image incuding the impexp and a Java JRE.
Currently, I manually delete several folders that are (apparently) not used by the ImporterExporter in the ``Dockerfile`` . 
See here: https://github.com/tum-gis/3dcitydb-importer-exporter-docker/blob/e17aff5002fc0310e0747dd51a1fbb8c669d3cf0/Dockerfile#L38

Would it be possible to add a gradle build task, that just builds the bare impexp cli version without example files usw.?
For instance ``./gradlew installCLI``, ``./gradlew distCLI``, ``./gradlew distCLIZIP``...
Assuming CLI users are experienced users, they probably don't require example files, the 3D webclient and so on. They just want the bare CLI.

impexp-client/src/main/java/org/citydb/cli/options/vis/DisplayOption.java
clausnagel: Yes, currently you must define the export format via `--display-mode`. Since the visibility setting is only optional (default: 0), I moved it to its own parameter `--visible-from`, which must therefore refer to the formats defined in `--display-mode`. You cannot add further display modes through `--visible-from`.

Both options could be merged into one: `display-mode=MODE[=VISIBLE_FROM],MODE[=VISIBLE_FROM]...`

For example:
```
--display-mode=collada // just export collada with visible from = 0
--display-mode=collada=50,footprint=100
```

The `picocli` library I m using does not support such a value list out-of-the-box, so it requires writing a specific option converter. I tried to avoid this with the current implementation.

Son-HNguyen: > Both options could be merged into one: ``display-mode=MODE[=VISIBLE_FROM],MODE[=VISIBLE_FROM]...``

This would be ideal for the next release(s) then.

BWibo:
#### Elevation settings for `export-vis`

I was just experiemnting with the `export-vis` using this set of options against the *lod3 railway dataset*.
* [CLI Opts](https://github.com/tum-gis/3dcitydb-importer-exporter-docker/blob/cli-rework/share/config/tests/export-vis-bldg.txt)
* Dataset: [tumgis/3dcitydb-postgis:railwayScene_LoD3](https://hub.docker.com/layers/tumgis/3dcitydb-postgis/railwayScene_LoD3/images/sha256-ecb8b2e663e7a8928c7b190d536854d7de4bd059354440fecc69969e7961beb7?context=explore)

I cound not find the options to control the elevation of the export. Am I missing something or is this not implemeted jet?
If no, I think it should be added. Those are frequently requried for visualization exports.
Unbenannt

#### One more thing regarding the `gltf options`
This works, `-O '--binary=true --darcoCompression=true'` but I could not get `-O 'b true'` running. 
I am getting this error: `Expected parameter for option '--gltf-option' but found '-b true'`. Am I doing something wrong here? What is the correct sntax for that case?

Son-HNguyen:
@BWibo Shouldn't it be called `dracoCompression`? And according to the [COLLADA2glTF converter](https://github.com/KhronosGroup/COLLADA2GLTF#options), you can just use `-b` for binary glTF and `-d` for Draco compression.

clausnagel:
Hi @BWibo, 

re 1: The elevation options are not available on the CLI yet. I wanted to start with a slim implementation. You can pass a config file to control the elevation settings. I therefore propose to create separate issues for including more CLI options after this PR has been merged.

re 2: The problem are the white spaces in your values for the `-O` option, and none of your alternatives is working in my tests. The reason is that the CLI parser uses white spaces to split the command line into tokens. You must use double quotes `"` to denote that the white spaces belong to your option value instead. This is true for all other options as well. Here are possible solutions to make your example work:

```
    -O "'--binary=true --dracoCompression=true'"
    -O '--binary=true' -O '--dracoCompression=true'
    -O '--binary=true','--dracoCompression=true'
    -O "'-b true -d true'"
    -O "'-b true'","'-d true'"
    ...
```

The reason why I had to introduce the single quote `'` is to make sure that an option like `--dracoCompression` is not interpreted as option for the `export-vis` command but basically treated as string. The leading dash (so either `--` or `-`) must be escaped for this purpose. We could also use a different escape character like `\` in order to avoid confusion between `'` and `"`. And note that the closing `'` is not really required - we could skip it as well.

So, with using `\`, the above examples could look like this:

```
    -O "\--binary=true --dracoCompression=true"
    -O \--binary=true -O \--dracoCompression=true
    -O \--binary=true,\--dracoCompression=true
    -O "\-b true -d true"
    -O "\-b true" -O "\-d true"
```

I do not really favor one alternative. Thoughts?

impexp-client/src/main/java/org/citydb/cli/options/vis/DisplayOption.java
clausnagel: I'd like to make this change **before** the next release, but outside this PR. The reason is that I reworked the config settings for KML/COLLADA/glTF exports in https://github.com/3dcitydb/importer-exporter/pull/149 because there was a lot of duplicate code, especially concerning the display forms. So implementing this proposal will be a lot easier after we have merged https://github.com/3dcitydb/importer-exporter/pull/149.

Son-HNguyen: Agreed.

BWibo:
@Son-HNguyen Yes, your right. I fixed the typo.

@clausnagel 
**re1:** OK, I'll do that when this is merged.
**re2:** Thx for clearing up the syntax. However, not all of your examples work for me, see below:
``` text
-O "'--binary=true --dracoCompression=true'"    => works
-O '--binary=true' -O '--dracoCompression=true' => works
-O '--binary=true','--dracoCompression=true'    => Unmatched arguments from index -1: ',', '--dracoCompression=true'
-O "'-b true -d true'"                          => works
-O "'-b true'","'-d true'"                      => Unmatched arguments from index -1: ',', ''-d true''
```

What is working as well is this: 
```text
-O "'-b','-d','-v 2.0'"
```
I think there is no need to required an escape character. However, I would include an example with the exact syntax in the CLI help if possible. You can easily get this wrong. I personally like the example syntax above best.

I've added some imports test in tum-gis/3dcitydb-importer-exporter-docker@9f532432e7033ccb05069e76afd4694300b96a97 and everytings seems fine.

From my side we can merge this.
Thank's for the great implementation @clausnagel, this is really helpful for automation of CityGML workflows.

As soon this is merged and released I'll finish the docker image and release it. By then we will have all core tools from the @3dcitydb dockerized! 🚀

BWibo (approved):

clausnagel:
Thanks for all your feedback and reviews. I will merge this PR now with the following open points to be addressed afterwards:

* merge `--display-mode` and `--visible-from` into one option
* add options for the elevation settings for KML/COLLADA/glTF exports
* clarify how to pass parameters to the collada2gltf converter (@BWibo do you have a suggestion for how to adapt the CLI help?)
* Discussion whether to support the `featureVersionFilter` filter

BWibo:
That's great!
Regarding the extension of the CLI help: Adding this example should clarify the usage for everyone:
`-O "'-b','--dracoCompression=true','-v 2.0'"`
It contains how multiple options are combined (`-O "'-opt val','-otheropt val'"`) and how the long (`--opt=val`) and short version (`-opt val`) of the different options of `collada2glTF` can be used:
```
-O, --gltf-option='<option>'[,'<option>'...]
                             CLI option to be passed to the COLLADA2GLTF
                             converter (embrace with single quotes). 
                             Embrace multiple options with double quotes, e.g.
                             -O "'-b','--dracoCompression=true','-v 2.0'"
```

clausnagel:
Thanks 👍 

Well, are you sure that you have pulled the latest version? Because all the examples work for me including `-O '--binary=true','--dracoCompression=true'` and `-O "'-b true'","'-d true'"`.

@yaozhihang, could you please test the examples on your machine, just to have a third opinion :-) Thanks.

BWibo:
I just created a fresh build (=inside docker image, impexp is build using gradle during image build) from 3dcitydb/importer-exporter@ad93127614ac42048e93445f07382d3fef7f5585 and ran the tests again.
* https://github.com/tum-gis/3dcitydb-importer-exporter-docker/blob/cli-rework/share/config/tests/export-vis/export-vis-gltfOpts-03.txt 

  ```
  # 3DCityDB ImporterExporter Docker #############################################
  ################################################################################
  #
  #   3DCityDB ImporterExporter version
  #     version info    https://github.com/3dcitydb/3dcitydb/tree/
  #
  # Maintainer -------------------------------------------------------------------
  #   Bruno Willenborg
  #   Chair of Geoinformatics
  #   Department of Aerospace and Geodesy
  #   Technical University of Munich (TUM)
  #   b.willenborg(at)tum.de
  #   www.gis.lrg.tum.de
  ################################################################################


  Command line passed to 3DCityDB ImporterExporter:
          @/share/config/tests/export-vis/export-vis-gltfOpts-03.txt -o /share/data/export/export-vis-gltfOpts-03/export-vis-gltfOpts-03.kml


  Unmatched arguments from index -1: ',', '--dracoCompression=true'
  Usage: impexp export-vis [-hjVz] [-c=<file>] [--log-file=<file>]
                          [--log-level=<level>] -o=<file> [--pid-file=<file>]
  ...
  ...                       
  ```

* https://github.com/tum-gis/3dcitydb-importer-exporter-docker/blob/cli-rework/share/config/tests/export-vis/export-vis-gltfOpts-05.txt

  ```
  # 3DCityDB ImporterExporter Docker #############################################
  ################################################################################
  #
  #   3DCityDB ImporterExporter version
  #     version info    https://github.com/3dcitydb/3dcitydb/tree/
  #
  # Maintainer -------------------------------------------------------------------
  #   Bruno Willenborg
  #   Chair of Geoinformatics
  #   Department of Aerospace and Geodesy
  #   Technical University of Munich (TUM)
  #   b.willenborg(at)tum.de
  #   www.gis.lrg.tum.de
  ################################################################################


  Command line passed to 3DCityDB ImporterExporter:
          @/share/config/tests/export-vis/export-vis-gltfOpts-05.txt -o /share/data/export/export-vis-gltfOpts-05/export-vis-gltfOpts-05.kml


  Unmatched arguments from index -1: ',', ''-d true''
  Usage: impexp export-vis [-hjVz] [-c=<file>] [--log-file=<file>]
                          [--log-level=<level>] -o=<file> [--pid-file=<file>]
                          [--worker-threads=<threads[,max]>] [-D=<mode>[,

  ...
  ...
  ```
I have no idea why this is working for you but not for me.
I am using the options from a `@file` using the docker image. Maybe this is related to Docker or the `@files`?

PR #148, opened 2020-11-19
enable github actions to test building

FxKu:
Just learned how easy GitHub actions can be recently. So maybe you like to have a build for each master commit and PR.


FxKu:
seems to work: https://github.com/3dcitydb/importer-exporter/actions/runs/372790382/workflow

yaozhihang:
Many thanks @FxKu. The PR looks good.

PR #149, opened 2020-11-29
New GUI Look and Feel

clausnagel:
This PR proposes a more modern Look and Feel for the Importer/Exporter GUI based on [FlatLaf](https://www.formdev.com/flatlaf/). I did not redesign the GUI but just "flattened" it a bit (for instance, by removing a lot or borders). Make sure to check out a really nice new feature which comes with FlatLaf - it is available from the "View" menu.

FlatLaf is cross-platform and scales on HiDPI displays. I tested the new LaF under Windows 10 and Ubuntu 20. 

Most of the commits in this PR are just about GUI changes. But some of them have a greater impact:
* I removed the `Workspace` dialog from all operation tabs. The workspace settings only make sense for Oracle but not for PostgreSQL. If we need them at all, I propose to integrate them on the `Database` tab as special fields just for Oracle databases. The workspace settings on the `Database` tab would then apply to all operations. 
* I rearranged the KML/COLLADA/glTF rendering preferences. There was a lot of duplicate code and, in fact, for many settings only the values entered for the `Building` feature were taken into account. I moved all general rendering options (like whether to use face normals, texture atlases, etc.) from the feature-specific settings to the `General` tab. This leaves only styling as feature-specific setting, which makes much more sense to me.

~~Note that this PR depends on https://github.com/3dcitydb/importer-exporter/pull/147. So, if you like the new LaF, we need to merge https://github.com/3dcitydb/importer-exporter/pull/147 first :-)~~ https://github.com/3dcitydb/importer-exporter/pull/147 has been merged.

If you use `gradlew installDist` for building a new Importer/Exporter for testing, make sure to first delete the `plugins` folder before starting the app. The plugins have not been moved to the new CLI yet and will throw errors preventing the Importer/Exporter from starting.


Son-HNguyen:
Oh yes

image

Son-HNguyen:
The window size in the default perspective fills almost the entire screen. The resolution of my screen is 2560x1440 and I would prefer the window size a bit smaller. I'm not sure how it looks on smaller/bigger resolutions though.

clausnagel:
Hi @Son-HNguyen, yes, this is mainly for lower resolutions so that you do not have to scroll in the separate operation tabs. I can use a minimum size instead so that the window would be smaller on your high resolution.

Son-HNguyen:
@clausnagel that would be great to adjust the window size with respect to the screen resolutions.

clausnagel:
Hm, I wonder whether we should also use FlatLaf for the installer. I don't want to rework the UI components themselves. But the installer does not scale well on my HiDPI display and is hardly readable. FlatLaf could help to fix this.

impexp-client/src/main/java/org/citydb/gui/modules/kml/preferences/GeneralPanel.java
```diff
@@ -464,6 +502,17 @@ private void setEnabledComponents() {
 		exportGltfV2.setEnabled(createGltfCheckbox.isSelected());
 		enableGltfDracoCompression.setEnabled(createGltfCheckbox.isSelected());
```
yaozhihang, line 503: here should be `enableGltfDracoCompression.setEnabled(createGltfCheckbox.isSelected() && exportGltfV2.isSelected());`

impexp-client/src/main/java/org/citydb/gui/modules/preferences/view/PreferencesPanel.java
```diff
@@ -242,11 +232,11 @@ public void doTranslation() {
...
 		standardButton.setText(Language.I18N.getString("common.button.standard"));
 		applyButton.setText(Language.I18N.getString("common.button.apply"));
-		noticeLabel.setText(Language.I18N.getString("common.pref.menu.expand.label"));
+		hintLabel.setText(Language.I18N.getString("common.pref.menu.expand.label"));
 
 		resetPreferencesMenu();
-		menuTree.repaint();
-		prefLabel.setText(((PreferencesTreeNode)menuTree.getLastSelectedPathComponent()).toString());	
+		settingsNameLabel.setText(menuTree.getLastSelectedPathComponent().toString());
+		splitPane.resetToPreferredSizes();
```
yaozhihang, line 239: `splitPane.repaint();` would be better. The `resetToPreferredSizes` sometimes crushes the left tree panel when switching preferences content panels.

clausnagel: What does "crushing" mean? I haven't experienced this so far. `resetToPreferredSizes()` is required for resizing, which `repaint()` doesn't do. So maybe adding `repaint()` helps to avoid the "crushing", but we need to keep `resetToPreferredSizes()`.

yaozhihang: For example: see the screenshot below (Note that there is a horizontal scroll bar at the bottom of the preferences content panel)
image
Now, if you click another tree node e.g. "gml:id Behandlung", and then click the "Fortschreibung" node again, the GUI will be shown like this: The tree panel has been _crushed_.
image

clausnagel: Does not seem to happen all the time though, only sometimes. Would be nice if you can find a solution that keeps `resetToPreferredSizes()` :-)

impexp-client/src/main/java/org/citydb/gui/modules/kml/preferences/GeneralPanel.java
clausnagel: fixed with 51e8d9c8bac3ac6292911183c419b0575cf727d8

Son-HNguyen: This is actually not needed because line 326 already takes care of it in runtime:
```
exportGltfV1.addItemListener(e -> enableGltfDracoCompression.setEnabled(!exportGltfV1.isSelected()));
```

Son-HNguyen (commented):
👍

impexp-client/src/main/java/org/citydb/gui/modules/preferences/view/PreferencesPanel.java
clausnagel: addressed in 1a3c7a284213268ff6895a9bcbf7d430b5ce159c

impexp-client/src/main/java/org/citydb/gui/modules/kml/preferences/GeneralPanel.java
clausnagel: @Son-HNguyen, well, without the fix, a situation like shown below is possible:

image

clausnagel: Still had issues with this checkbox and hope that d5c74a562df3df06cb6fe3275c34abd519cf5201 now fixes them. Please check :-)

clausnagel:
> [e24c89b](https://github.com/3dcitydb/importer-exporter/commit/e24c89b5a7e21058ebd6954c789fe8f554442632) 👍

Does it work for you? 

While working on it, I realized that Java itself has still bugs especially if you use multiple screens with different scale factors (like in my environment). For instance, moving the window from one screen to the other, detaching the console or restoring the default layout can result in strange situations in such settings. I tried to find workarounds but have now decided to keep the current implementation as is and wait for a bugfix in Java.

Son-HNguyen:
It does work for me with a dual monitor setup, but both have the same resolution of 2560x1440.

yaozhihang (approved):

yaozhihang:
@clausnagel create work! from my side, this PR is OK for merge.

Issue #150, opened 2020-12-03
Rework CLI options for KML/COLLADA/glTF export


clausnagel:
The following points have been raised in the discussion of #147:

- [x]  merge `--display-mode` and `--visible-from` into one option
- [x] add options for the elevation settings for KML/COLLADA/glTF exports
- [x] add explicit CLI options for the collada2gltf converter and skip `-O`
was: ~~clarify how to pass parameters to the collada2gltf converter~~

Check out the `export-vis-cli` branch to follow and test the changes.

clausnagel:
@BWibo, I have checked your issue with passing parameters to the collada2gltf converter using @-files. I can confirm that some notations like, for example, `-O '--binary=true','--dracoCompression=true'` work on the CLI but not when using @-files. This example however worked for me after omitting the single quotes `'` in the @-file.

Well, this is rather confusing for users. So, my proposal is to drop the `-O` option and to define specific options like `--gltf-version` and `--gltf-draco-compression`. Opinions?

clausnagel:
Ok, just learned that the parsing of option names _should_ just work fine in picocli as illustrated in the [online documentation](https://picocli.info/#_option_names_as_option_values). But it does not work. Seems to be a bug in picocli. Just checking whether I can work around the bug, so that `-O` can be kept.

yaozhihang:
I like your proposal of dropping the `-O` option and introducing three options `--gltf-version`, `--gltf-binary`, and  `--gltf-draco-compression` instead. This makes the CLI interface clearer and easier to understand.

clausnagel:
I merged the `--visible-from` option into the `--display-mode` option.

I removed the `-O` option and added `--gltf-version`, `--gltf-embed-textures`, `--gltf-binary` and `--gltf-draco-compression` as explicit options to control the behaviour of the collada2gltf converter. Note that the CLI supports abbreviated option names as explained in the [picocli online documentation](https://picocli.info/#_abbreviated_options_and_subcommands). Thus, for example, you don't have to type the full option name `--gltf-draco-compression` but can also use `--gDC`. Much simpler :-) This works for all options of all commands.

I also added options to control the elevation and mainly followed the proposal of @Son-HNguyen in #139 for this.

The `export-vis` is now a monster command (just print out the help). But, well, the feedback shows we obviously need all those options. So, I am thinking to also add the COLLADA rendering options as proposed by @Son-HNguyen in #139 (see `[rendering_options]`). 

[...]

PR #151, opened 2020-12-07
Add more options to export-vis command

clausnagel:
See #150


impexp-config/src/main/java/org/citydb/config/project/kmlExporter/GltfOptions.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public GltfOptions() {
+        gltfVersion = GltfVersion.v2_0;
+        embedTextures = true;
+        useDracoCompression = true;
+
+        pathToConverter = "contribs" + File.separator + "collada2gltf";
+        String osName = System.getProperty("os.name").toLowerCase(Locale.ENGLISH);
+        if (osName.contains("windows"))
+            pathToConverter += File.separator + "COLLADA2GLTF-v2.1.3-windows-Release-x64" + File.separator + "COLLADA2GLTF-bin.exe";
```
BWibo, line 61: v 2.15 of COLLADA2glTF is available: https://github.com/KhronosGroup/COLLADA2GLTF/releases/tag/v2.1.5

BWibo (commented):
For high res textures I sometimes use `textureImageScaleFactor` to down-scale texture images to increase visualization performance. 
`WallFillColor` and `RooFillfColor` are frequently used as well for styling or e.g. to create thematic visualizations in web client demos. It would be even cooler to be able to set these colors from a `GenericAttribute` or even map a `ColorPalette` to a value range. But this goes too far for this PR. 
Being able to set these values from the CLI and have them with the options below would already be great.

```
COLLADA/glTF rendering options:
  -s, --double-sided         Force all surfaces to be double sided.
      --no-surface-normals   Do not generate surface normals.
  -C, --crop-textures        Crop texture images.
  -x, --texture-atlas=<mode> Texture atlas mode: none, basic, tpim,
                               tpim_wo_rotation (default: basic).
      --no-pot-atlases       Do not create power-of-two sized texture atlase
```

I added some tests for the `AltitudeOptions` in https://github.com/tum-gis/3dcitydb-importer-exporter-docker/commit/f23a9824da77d13a60989a34443cb925c73e33e8 and it appears to be working fine. All other tests look good too.

impexp-config/src/main/java/org/citydb/config/project/kmlExporter/GltfOptions.java
clausnagel: Yes, we have already tested with version 2.1.5 but it fails when using Draco compression. 2.1.4 seems to work though. We will have to do more tests to confirm.

clausnagel:
Thanks for your continuous tests, @BWibo. This is very helpful.

I will add an option to set the image scale factor. Styling options are definetly out of scope of this PR though. You can define the styling in a config file and pass it via the `--config` option. It does not even have to be a full config but can be tailored to the styling content as shown below.

```xml
<kmlExport xmlns="http://www.3dcitydb.org/importer-exporter/config">
  <buildingStyles>
    <style type="collada">
    ...
    </style>
    ...
  </buildingStyles>
</kmlExport>
```

In general, I share your wish to be able to have more styling options like styling based on attributes and/or color ranges. I also think that we already should change the current styling options to be able to define colors for _all_ feature types. @Son-HNguyen, @thomashkolbe, maybe something you can take a look at?

clausnagel:
Ok, I added the possibility that the `--bbox-tiling` can take a tile side length for automatic tiling (see #150).

I see this as last change to the `export-vis` command and want to merge this PR if there is no further feedback on the existing options.

impexp-config/src/main/java/org/citydb/config/project/kmlExporter/GltfOptions.java
clausnagel: Updated COLLADA2glTF converter to v2.1.4 with 89a163d6326dc4d03e5cf6345b8e67f41338d60d. Further tests confirmed that version 2.1.5 has issues when using Draco compression.

yaozhihang:
@clausnagel thanks for the nice work. The PR looks good to me. 
However, there is sitll an small issue regarding the query option. The current KML/Collada/glTF Exporter requires at least a gml:id filter or a bounding box filter. However, the CLI tells that the query option is optional. To solve this problem, we may need to implement the automatic calculation of bounding box. This should be done in a separate MR.

yaozhihang (approved):

BWibo:
I added tests for `-f` and `-g auto=tileLength` in https://github.com/tum-gis/3dcitydb-importer-exporter-docker/commit/379ce1a34273f4216ab7dcccb0067216fda6a05a and everything seems fine.
Thx for the great work @clausnagel.

BWibo (approved):

clausnagel:
@yaozhihang: Yes, `--bbox` and `--gml-id` are optional because the query settings can also be provided in a config file through the `--config` option. We still want to support the "old" way of using the CLI so that you can provide a "full" config file containing all export options. 

If you run an export without providing a bbox or gml:id list one or the other way, the export will fail with a corresponding error message.

I agree that the automatic calculation of the bbox is a much smarter solution, and we should address the topic soon.

Issue #138, opened 2020-09-18
Automatically calculate the BBox for KML/COLLADA/glTF exports


clausnagel:
A KML/COLLADA/glTF export requires to either provide a list of gml:ids or a BBox. Both inputs should rather be optional and if omitted, the exporter should calculate the required BBox automatically in the background. This should even be the default option. If additional export filters are provided, the BBox should only be calculated based on the features matching the filters. 

See also the discussion in PR #96.

yaozhihang:
Completed in this PR https://github.com/3dcitydb/importer-exporter/pull/153.

PR #153, opened 2020-12-11
Support tiled vis-export with non-spatial filters

yaozhihang:
This PR solves the issue https://github.com/3dcitydb/importer-exporter/issues/138
When the bbox filter is disabled, the BBox can now be automatically calculated based on the features matching the other filters. In addition, tiling can also be applied based on the calculated BBOX optionally.


impexp-core/src/main/java/org/citydb/database/adapter/AbstractUtilAdapter.java
```diff
@@ -184,6 +188,48 @@ public BoundingBox calcBoundingBox(Workspace workspace, List<Integer> objectClas
...
+                            new Position(coordinates[0], coordinates[1]),
+                            new Position(coordinates[2], coordinates[3])
+                        );
+                        bbox.setSrs(extent.getSrid());
+
+                        DatabaseSrs targetSrs = query.getTargetSrs();
+                        if (targetSrs != null && bbox.getSrs().getSrid() != targetSrs.getSrid()) {
+                            Tiling tiling = new Tiling(bbox, 1, 1);
+                            tiling.transformExtent(targetSrs, databaseAdapter);
+                            bbox = tiling.getExtent();
```
clausnagel, line 223: Creating a `Tiling` object just to transform the bbox looks like unecessary overhead. Shouldn't we rather extract just those lines dealing with the transformation from the `Tiling.transformExtent` method and put them in a new method here in the util adapter? Then both this method and  `Tiling.transformExtent` could reuse it. Maybe this new method could even replace the depracted `transformBoundingBox` method in the util adapter?

impexp-core/src/main/java/org/citydb/query/builder/config/ConfigQueryBuilder.java
```diff
@@ -317,19 +319,52 @@ public Query buildQuery(SimpleKmlQuery queryConfig, NamespaceContext namespaceCo
...
+					extentQuery.setSelection(new SelectionFilter(predicates.get(0)));
+				}
+				spatialFilter.setExtent(databaseAdapter.getUtil().calcBoundingBox(workspace, query, schemaMapping));
+			} catch (SQLException | FilterException e) {
+				throw new QueryBuildException("Failed to calculate bounding box based on the non-spatial filter settings.", e);
+			}
+		}
+
+		// bbox filter
+		if (queryConfig.isUseBboxFilter()) {
```
clausnagel, line 345: l. 345: else if?!

impexp-core/src/main/java/org/citydb/query/builder/config/ConfigQueryBuilder.java
```diff
@@ -317,19 +319,52 @@ public Query buildQuery(SimpleKmlQuery queryConfig, NamespaceContext namespaceCo
...
+			}
+		}
+
+		// apply tiling
+		if (spatialFilter.getMode() != KmlTilingMode.NO_TILING && spatialFilter.isSetExtent()) {
+			TilingFilterBuilder tilingFilterBuilder = new TilingFilterBuilder(databaseAdapter);
+			query.setTiling(tilingFilterBuilder.buildTilingFilter(spatialFilter));
+		}
+
+		if (!predicates.isEmpty()) {
```
clausnagel, line 362: If neither gml:id, bbox nor tiling is set, then still an extent is calculated in line 338 (because `!queryConfig.isUseBboxFilter()` is true) and the result is assigned to the `spatialFilter` object. But this extent is not used anywhere later, because `spatialFilter.getMode() != KmlTilingMode.NO_TILING` is false in line 357. So, the extent is calculated but never used in this scenario. I think this needs to be adapted here.

What is the intended purpose of the automatic bbox? Should it always be calculated and used whenever the user does not provide a bbox?

impexp-kml-collada-plugin/src/main/java/org/citydb/modules/kml/controller/KmlExporter.java
```diff
@@ -305,29 +247,29 @@ else if (!Files.isExecutable(collada2gltf))
...
-				// transform tiling extent to WGS84
-				tiling.transformExtent(DatabaseConfig.PREDEFINED_SRS.get(DatabaseConfig.PredefinedSrsName.WGS84_2D), databaseAdapter);
-				tilingOptions = tiling.getTilingOptions() instanceof KmlTilingOptions ? (KmlTilingOptions)tiling.getTilingOptions() : new KmlTilingOptions();
-				predicate = query.isSetSelection() ? query.getSelection().getPredicate() : null;
-
-				// calculate and display number of tiles to be exported
-				int displayFormats = config.getKmlExportConfig().getDisplayForms().getActiveDisplayFormsAmount();
-				remainingTiles = rows * columns * displayFormats;
-				log.info(remainingTiles + " (" + rows + "x" + columns + "x" + displayFormats + ") tiles will be generated.");	
+				tiling = new Tiling(config.getKmlExportConfig().getQuery().getSpatialFilter().getExtent(), rows, columns);
```
clausnagel, line 255: I am not sure, but it seems that this line assumes that the spatial extent was automatically calculated in the `ConfigQueryBuilder`. This would explain why the spatial extent is calculated there even if it is not used (see my comment above).

If this is true, I recommend to resole this "side effect dependency". I suggest that only the `KMLExporter` controller should check whether or not a bbox must be automatically calculated based on the provided query config. And it should trigger the calculation of the bbox, and adapt the query config afterwards if needed. This (possibly adapted) query config should then be passed to the `ConfigQueryBuilder`, and the `ConfigQueryBuilder` should just map it to `Query` object (no further logic).

impexp-client/src/main/java/org/citydb/gui/modules/kml/view/KmlExportPanel.java
```diff
@@ -377,10 +367,10 @@ private void updateComponentUI() {
 
     public void doTranslation() {
         browseButton.setText(Language.I18N.getString("common.button.browse"));
-        gmlIdPanel.setTitle(Language.I18N.getString("kmlExport.label.singleBuilding"));
+        gmlIdPanel.setTitle(Language.I18N.getString("filter.border.attributes"));
```
clausnagel, line 370: The keys "kmlExport.label.singleBuilding" and "kmlExport.label.noTiling" can be deleted from the language files.

impexp-core/src/main/java/org/citydb/database/adapter/AbstractUtilAdapter.java
yaozhihang: Yes, agree. Maybe we should make this small refactorying in a separate dedicated PR?

impexp-core/src/main/java/org/citydb/query/builder/config/ConfigQueryBuilder.java
yaozhihang: Since the output master KML and JSON files require the extent of the exported features, a bbox must be either automatically calculated or provided by users.

impexp-kml-collada-plugin/src/main/java/org/citydb/modules/kml/controller/KmlExporter.java
yaozhihang: You are right, line 255 assumes that the spatial extent was automatically calculated. However, it is indeed used in the output master and JSON files.

impexp-core/src/main/java/org/citydb/database/adapter/AbstractUtilAdapter.java
clausnagel: Agreed

impexp-core/src/main/java/org/citydb/query/builder/config/ConfigQueryBuilder.java
clausnagel: See below

impexp-kml-collada-plugin/src/main/java/org/citydb/modules/kml/controller/KmlExporter.java
clausnagel: Yes, understood. Testing and automatically calculating the bbox should better be done in `KMLExporter` and not in `ConfigQueryBuilder` in my opinion. This would be clearer and avoid the side effect.

yaozhihang: Good idea. Done in 4327dbc

impexp-client/src/main/java/org/citydb/gui/modules/kml/view/KmlExportPanel.java
yaozhihang: Done here 4016fcb

impexp-core/src/main/java/org/citydb/query/builder/config/ConfigQueryBuilder.java
yaozhihang: Resolved here 4327dbc

clausnagel:
The following simple CLI command leads to an error:

```
export-vis -H localhost -d devel -u cnagel -o C:\data\kml\test.kml -D geometry -l halod
```

```
[10:32:56 INFO] Checking for spatial indexes on geometry columns of involved tables...
[10:32:56 ERROR] Failed to build the internal fallback tiling.
[10:32:56 ERROR] Caused by: org.citydb.query.filter.FilterException: The bounding box extent is invalid.
[10:32:56 WARN] Database export aborted.
```

The problem seems to be that when creating a `KmlExportConfig` also a `SimplqKmlQuery` is automatically instantiated, and this `SimplqKmlQuery` contains a default `KmlTiling` object with an empty bbox. This empty bbox is created in the constructor of `AbstractTiling` which is the root for CityGML and KML tiling.

Now, since my CLI command does not contain query options, your current code in `org.citydb.cli.options.vis.QueryOption` to unset the bbox (see line 75 there) is never executed.

There are different possibilities to solve this. You could add another test in `ExportVisCommand` to unset the bbox if no query options have been provided. But this does not feel correct (too many tests for edge cases at too many places).

I think it's better to not create an empty bbox per default in the constructor of `AbstractTiling`. Then the issue would be solved for KML/COLLADA/glTF exports. Since the CityGML export (still) requires a bbox, the creation of the empty bbox must be moved into the constructor of `org.citydb.config.project.exporter.SimpleTiling` to avoid side effects at this moment. This way, the bbox must not be explicitly set to null in `org.citydb.cli.options.vis.QueryOption` anymore and line 75 of this class can be removed. So we get rid of egde case tests.

I like the second option much better. I hope I could explain it clearly? If not, I can change the code correspondingly in this PR if you want me to.

yaozhihang:
@clausnagel Good catch, thanks for the explanation. The second option looks much better. Based on it, I would like to propose a third option (see the commit 0aa7520). This way, there is no need to change the three classes `org.citydb.config.project.query.filter.tiling.AbstractTiling`, `org.citydb.config.project.exporter.SimpleTiling`, and `org.citydb.config.project.query.filter.tiling.Tiling`.

clausnagel:
Thanks @yaozhihang, agreed. We should revisit this when we add automatic bbox calculation to the CityGML export.

clausnagel (approved):

clausnagel:
Final question just for information: As I understand, every export is now a tiled export, even if it's just one tile. This means that there will always be one master KML file (the file that the user specifies on the `KML/COLLADA/glTF export` tab), right? And the "Export as KMZ" option only applies to the separate tiles, not to the master KML file, right? So, with this PR, there is no need to be able to specify a KMZ file on the `KML/COLLADA/glTF export` tab anymore?

If this is true, this would allow us to tidy up some code in the KML/COLLADA/glTF exporter. But this work does not affect this PR.

BWibo (approved):

yaozhihang:
@clausnagel Yes, that's totally right. 
Earlier, the vis-exports with gml:id filter are mapped to a single KML or KMZ file. No master json file is genereated and can therefore not be handled by the 3ditydb-web-map using its tiling mechanism. This limitation has now been resolved in this PR, since all vis-exports are now mapped to the tiling structure. Yes, a single tile will be created, if the tiling option is not applied.

PR #155, opened 2020-12-15
added CLI options --plugins and --ade-extensions

yaozhihang:


clausnagel:
picocli 4.6.0 introduced a new preprocessor parser plugin called `IParameterPreprocessor`. This might be an option to dynamically load CLI plugins from a provided folder as proposed in this PR but still to use the parsing capabilities of picocli. I will take a look at it and re-open this PR if it works.

clausnagel:
Reopened this PR.

Unfortunately, the new picocli preprocessor does not really help. But what I did is to adapt the original proposal as follows:

1. The CLI generally supports abbreviated forms for all options. For instance, instead of `--plugins` you can also provide `--plu` as long as the abbreviated form is unique. I added this behaviour to the parsing of the `--plugins` option.
2. Folders provided through `--plugins` and `--ade-extensions` are resolved against the current working directory (`ClientConstants.WORKING_DIR`) instead of the installation directory (`ClientConstants.IMPEXP_HOME`). 

@yaozhihang, would you please review?

yaozhihang:
@clausnagel thanks very for the revision. It looks very good and works well in my tests.

PR #156, opened 2020-12-17
move FilterView classes to common package

yaozhihang:
This PR uncouples the FilterView APIs from the exporter package, so that the APIs can be used in external modules e.g. plugins.

### DIFF
--- a/impexp-client/src/main/java/org/citydb/gui/modules/common/filter/XMLQueryView.java
+++ b/impexp-client/src/main/java/org/citydb/gui/modules/common/filter/XMLQueryView.java
@@ -26,11 +26,10 @@
  * limitations under the License.
  */
 
-package org.citydb.gui.modules.exporter.view.filter;
+package org.citydb.gui.modules.common.filter;
 
 import com.formdev.flatlaf.extras.FlatSVGIcon;
 import com.sun.xml.bind.marshaller.NamespacePrefixMapper;
-import org.citydb.config.Config;
 import org.citydb.config.ConfigUtil;
 import org.citydb.config.geometry.BoundingBox;
 import org.citydb.config.i18n.Language;
@@ -59,13 +58,13 @@ import org.citydb.config.project.query.filter.type.FeatureTypeFilter;
 import org.citydb.config.project.query.simple.SimpleFeatureVersionFilter;
 import org.citydb.config.project.query.simple.SimpleFeatureVersionFilterMode;
 import org.citydb.config.project.query.simple.SimpleSelectionFilter;
+import org.citydb.config.util.ConfigNamespaceFilter;
 import org.citydb.config.util.QueryWrapper;
 import org.citydb.database.connection.DatabaseConnectionPool;
 import org.citydb.database.schema.mapping.FeatureType;
 import org.citydb.database.schema.mapping.SchemaMapping;
 import org.citydb.gui.factory.PopupMenuDecorator;
 import org.citydb.gui.factory.RSyntaxTextAreaHelper;
-import org.citydb.gui.modules.exporter.view.FilterPanel;
 import org.citydb.gui.util.GuiUtil;
 import org.citydb.log.Logger;
 import org.citydb.plugin.extension.view.ViewController;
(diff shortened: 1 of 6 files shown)


impexp-client/src/main/java/org/citydb/gui/modules/common/filter/XMLQueryView.java
```diff
@@ -463,14 +467,22 @@ public Icon getIcon() {
...
-        config.getExportConfig().setQuery(query);
+        queryConfigSupplier.get().setTargetSrs(query.getTargetSrs());
+        queryConfigSupplier.get().setFeatureTypeFilter(query.getFeatureTypeFilter());
+        queryConfigSupplier.get().setAppearanceFilter(query.getAppearanceFilter());
+        queryConfigSupplier.get().setCounterFilter(query.getCounterFilter());
+        queryConfigSupplier.get().setLodFilter(query.getLodFilter());
+        queryConfigSupplier.get().setProjectionFilter(query.getProjectionFilter());
+        queryConfigSupplier.get().setSelectionFilter(query.getSelectionFilter());
+        queryConfigSupplier.get().setSorting(query.getSorting());
+        queryConfigSupplier.get().setTiling(query.getTiling());
```
clausnagel, line 485: Maybe we should add a `copyFrom(QueryConfig other)` to the `QueryConfig` class? Then we could replace lines 477-485 with one call `queryConfigSupplier.get().copyFrom(query)`.

The advantage is that the `QueryConfig` class must make sure that all elements of a query are correctly copied. For example, if we extend the `QueryConfig` class in future, we only need to adapt the `copyFrom` method to account for these changes.

yaozhihang: Good idea

impexp-client/src/main/java/org/citydb/gui/modules/common/filter/XMLQueryView.java
```diff
@@ -463,14 +467,22 @@ public Icon getIcon() {
 
     @Override
     public void loadSettings() {
-        QueryConfig query = config.getExportConfig().getQuery();
-        xmlText.setText(marshalQuery(query, config.getNamespaceFilter()));
+        QueryConfig query = queryConfigSupplier.get();
+        xmlText.setText(marshalQuery(query, configNamespaceFilterSupplier.get()));
```
clausnagel, line 471: I think we should only require suppliers for settings that are specific to the "user" of the filter view (for instance, "user" = export or delete operation).

The `ConfigNamespaceFilter` is not user-specific. It is created when loading the `Config` and only changes when loading a new config. So, I recommend to use `ConfigNamespaceFilter namespaceFilter = ObjectRegistry.getInstance().getConfig().getNamespaceFilter();` instead.

yaozhihang: Agree. Done in commit cec12f3

impexp-client/src/main/java/org/citydb/gui/modules/common/filter/XMLQueryView.java
yaozhihang: Done

clausnagel: Thanks. Well, this change was indeed required. Otherwise, this line `query.setLocalProperty("unmarshallingFailed", true);` (line 383) would have been without effect. It now works again because your `copyFrom` method also takes care of the local properties.

Hm, instead of copying we could also require another `@FunctionalInterface` in the constructor which sets the query object. So instead of

`queryConfigSupplier.get().copyFrom(query)`

something like

`queryConfigSetter.set(query)`

Then we take the object as created and avoid further side effects.

yaozhihang: Yes, this approach looks much cleaner. see here 372d897

clausnagel (approved):

Issue #154, opened 2020-12-12
Resolve the deprecated `transformBoundingBox` method in the UtilAdapter class


yaozhihang:
This method picks the lowerLeft and upperRight corners of the input bbox as stub for building the output bbox. Obviously, the result is not the _minimum bounding box_ (MBB) of the input bbox due to the distortion caused by CRS transformation (e.g. projected CRS <-> geographic CRS). In order to obtain MBB, the method `transformExtent` in the class `org.citydb.query.filter.tiling.Tiling` shall be used. We should consider moving this `transformExtent` method into the util adapter and replace the deprecated `transformBoundingBox` method.

clausnagel:
solved with #157.

PR #157, opened 2020-12-17
Resolve deprecated transformBoundingBox method

clausnagel:
This PR solves #154

* Removed deprecated `transformBoundingBox` method
* Harmonized the transformation of bounding boxes: 1) the bounding box is converted into a polygon, 2) the polygon is transformed, 3) the extent of the transformed polygon is returned as new bounding box.

### DIFF
--- a/impexp-config/src/main/java/org/citydb/config/geometry/GeometryObject.java
+++ b/impexp-config/src/main/java/org/citydb/config/geometry/GeometryObject.java
@@ -291,16 +291,20 @@ public class GeometryObject {
     }
 
     public GeometryObject toEnvelope() {
-        if (geometryType == GeometryType.ENVELOPE) {
-            return this;
-        }
-
         GeometryObject envelope = new GeometryObject(GeometryType.ENVELOPE, dimension, srid);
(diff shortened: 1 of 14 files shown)


impexp-config/src/main/java/org/citydb/config/geometry/GeometryObject.java
```diff
@@ -291,6 +291,10 @@ public void changeSrid(int srid) {
     }
 
     public GeometryObject toEnvelope() {
+        if (geometryType == GeometryType.ENVELOPE) {
+            return this;
```
yaozhihang, line 295: `return this` may lead to an inconsistent behaviour of this function, because a new envelope object is expected to be returned.

clausnagel: You are right, will adapt this.

clausnagel: see here 474aa1acb54f8a6850eb9b46ae5ca1d87a6f0a01

yaozhihang (approved):

PR #158, opened 2020-12-19
added support for previewing delete process

yaozhihang:


impexp-client/src/main/java/org/citydb/cli/DeleteCommand.java
```diff
@@ -54,6 +54,9 @@
             description = "Delete mode: ${COMPLETION-CANDIDATES} (default: ${DEFAULT-VALUE}).")
     private Mode mode;
 
+    @CommandLine.Option(names = "--preview", description = "Only print a report of the matched top-level city objects.")
```
clausnagel, line 57: I think it would be nice if there would also be a one-character option name in addition to `--preview`. Both `-p` and `-P` are already in use by other options, but what about `v`?

impexp-client/src/main/java/org/citydb/cli/DeleteCommand.java
```diff
@@ -90,20 +94,20 @@ public Integer call() throws Exception {
...
                 if (queryOption != null) {
                     config.getDeleteConfig().setQuery(queryOption.toQueryConfig());
                 }
 
-                deleter.doDelete();
+                deleter.doDelete(preview);
             }
 
-            log.info("Database delete successfully finished.");
+            log.info("Database " + operation + " successfully finished.");
```
clausnagel, line 107: I think `"Database preview successfully finished."` and also `"Database preview aborted."` are not really meaningful. Also other operations might have a preview mode so that you cannot say what sort of preview finished or aborted. I would rather prefer the operation name to be either `delete` or `terminate`.

That the delete/terminate operation is running in a preview mode should rather be explicitly printed to the log as a general `info` message. I think this info message is missing in the current implementation. It should make clear that the process will only check which top-level features would be affected by a delete/terminate but that the feature will not be deleted/terminated.

clausnagel:
Thanks @yaozhihang, only two minor comments from me (see above)

impexp-client/src/main/java/org/citydb/cli/DeleteCommand.java
yaozhihang: yes, see here c183afb

impexp-client/src/main/java/org/citydb/cli/DeleteCommand.java
yaozhihang: Agree, Done here dcd650e

clausnagel: Thanks, @yaozhihang. I think the info message must be placed in the `Deleter` controller instead because the preview mode can also be invoked through the GUI or programmatically. I have attached a patch file which you may want to use.

[info-message.zip](https://github.com/3dcitydb/importer-exporter/files/5728770/info-message.zip)

yaozhihang: @clausnagel thanks very much for the patch. It is used here 0e6b426

clausnagel (approved):

PR #162, opened 2021-01-04
PostGIS as default database connection

b-schwab:
As far as I am aware, the majority use the Importer/Exporter for a PostgreSQL/PostGIS database solution.
Could the PostGIS option in the GUI be the default selection for a new database connection?


clausnagel:
Thanks for your PR, @b-schwab. I don't know whether your assumption is correct but I have no objection to changing the default database for new connections :-) I am not sure but there might be more places in the code where Oracle is used as default. 

But before looking into this, are there any other opinions?

BWibo:
I fully agree to switch to PostgreSQL/PostGIS as default setting.
The change does no harm anybody and I assume as well that the PostgreSQL/PostGIS user base is much larger.

clausnagel:
Ok, will merge this now. The CLI already uses PostgreSQL as default. I will do one more change to the `org.citydb.config.project.database.DatabaseType` enum after merging.

PR #163, opened 2021-01-07
added support for creating minimized distribution contents

yaozhihang:
With this PR, a new gradle task `installMinimizedDist` is provided for generating minimized distribution contents, which are mandatorily required for running the importer-exporter. Non-mandatory folders like `3dcitydb`, `samples`, `templates`, and `3d-web-map-client` are omitted. This new task is especially useful when running the importer-exporter via Docker to keep the Docker image small.

### DIFF
--- a/impexp-client/build.gradle
+++ b/impexp-client/build.gradle
@@ -141,6 +141,21 @@ distributions.main {
     }
 }
 
+distributions {
+    minimized {
+        distributionBaseName = project.appName + '-Minimized'
+        contents {
+            with distributions.main.contents
+            exclude {
+                it.getFile().getParentFile().compareTo(file("$rootDir/resources/3dcitydb")) == 0 ||
+                it.getFile().getParentFile().compareTo(file("$rootDir/resources/samples")) == 0 ||
+                it.getFile().getParentFile().compareTo(file("$rootDir/resources/templates")) == 0 ||
+                it.getFile().getParentFile().compareTo(file("$rootDir/resources/3d-web-map-client")) == 0
+            }
+        }
+    }
+}
+
 task buildInstaller(dependsOn: installDist, group: 'distribution') {
     def installDir = installDist.destinationDir.toString()
     def distDir = distZip.destinationDirectory.get()
--- a/impexp-client/build.gradle
+++ b/impexp-client/build.gradle
@@ -143,7 +143,7 @@ distributions.main {
 
 distributions {
     minimized {
-        distributionBaseName = project.appName + '-Minimized'
+        distributionBaseName = project.appName
         contents {
             with distributions.main.contents
             exclude {
--- a/impexp-client/build.gradle
+++ b/impexp-client/build.gradle
@@ -142,11 +142,13 @@ distributions {
         }
     }
 
-    minimized {
-        distributionBaseName = project.appName + '-Minimized'
+    docker {
+        distributionBaseName = project.appName + '-Docker'
         contents {
             with distributions.main.contents
+            exclude ('*.bat', 'collada2gltf/*osx*', 'collada2gltf/*windows*')
             exclude {
+                it.getFile().compareTo(file("$rootDir/resources/plugins/plugin-ade-manager")) == 0 ||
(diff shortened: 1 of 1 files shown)


impexp-client/build.gradle
```diff
@@ -141,6 +141,21 @@ distributions.main {
     }
 }
 
+distributions {
+    minimized {
+        distributionBaseName = project.appName + '-Minimized'
```
BWibo: I suggest to leave `project.appName` as it is. This changes the output folder and therefore breaks the `Dockerfiles`.
If there is a good reason the the changed naming it not a big deal to change this in the `Dockerfiles`though.

BWibo:
Anything else seems fine, thanks for the implementation @yaozhihang. Currently, I delete `/impexp/contribs/collada2gltf/*osx*` and `/impexp/contribs/collada2gltf/*windows*` as they are not needed on for the UNIX-based images and some other stuff in the `Dockerfiles`, but it seems not worth to implement this with gradle to me.

BWibo (approved):

impexp-client/build.gradle
```diff
@@ -141,6 +141,21 @@ distributions.main {
     }
 }
 
+distributions {
+    minimized {
```
clausnagel, line 145: Not really an issue, but I suggest to only have one `distributions` property which then contains both `main` and `minimized`:

```gradle
distributions {
  main {
    ...
  }

  minimized {
    ...
  }
}
```

clausnagel:
@BWibo, yes, I was also thinking about the COLLADA2GLTF files. For a Docker image, we only need the Linux version of the tool and can drop the Windows/Mac versions, which helps to greatly reduce the size of the resulting Docker image. If the only purpose of the `minimzed` installation is to generate a Docker image, then we could implement this in the gradle script.

But I think we should keep the `minimzed` installation more general so that it can also be used on other platforms. This means that the Windows/Mac versions of COLLADA2GLTF indeed have to be removed by the Dockerfile script and not the gradle script.

BWibo:
Yes, I think so too. We would lose cross-platfrom capabilities if we removed the COLLADA2GLTF dists for OSX and WINDOWS for the `minimized` version. Moreover, deleting this in the `Dockerfile` is one line of code that is probably not going to be subject of frequent future change.

BWibo:
In 1dc36253a8757da832603c298e55141ba3c3296d I moved the `minimized` dist into the `distributions` property as discussed aboved and changed `distributionBaseName` back to `distributionBaseName = project.appName + '-Minimized'` to avoid confliction dist folders. I'll adapt the paths in the `Dockerfiles`.

yaozhihang:
Thanks very much for all your inputs @BWibo @clausnagel. The current implementation looks good to me.  Maybe we could think about adding three additional tasks (see the proposal below)? This allows for building OS-specific distributions much easier. Also, it will make the `Dockerfile` simpler and better readable. Agree?

```gradle
distributions {
    main {
        ...
    }

    minimized {
        ...
    }

    minimized4Windows {
        ...
    }

    minimized4MacOS {
        ...
    }

    minimized4Linux {
        ...
    }
}
```

clausnagel:
I don't think that we need additional, OS-specific minimized installation tasks. It should be fairly simply to remove the OS-specific files of the COLLADA2GLTF converter in a `Dockerfile`. For end-users, our installer already takes care of OS-specific installations. And if we would introduce, for example, a `minimized4Windows`, wouldn't we have to offer a `full4Windows` also?

I would rather like to turn the question around as in my previous post: do we really need the `minimized` task in its current form? My understanding is that it's main purpose is to make the automatic creation of a Docker image easier. Then why not call it `docker` task and make sure that this task creates an installation that is optimized for use with Docker?

yaozhihang:
@clausnagel you are right, the initial purpose of this PR is indeed for facilitating the creation of Docker image. I agree to rename the task `minimized` to `docker` to make things straightforward. Please check the commit b4b414f which renames the task and further optimzes the size of the distribution contents for use with Docker.

clausnagel:
Thanks, @yaozhihang. For me, it looks good now.

clausnagel (approved):

PR #164, opened 2021-01-08
Workspace settings for Oracle databases

clausnagel:
This PR adds GUI elements to enter the target workspace for Oracle databases on the `Database` tab. The config file has been reworked to store one workspace per connection (and not one workspace per operation as before). This also has an impact on various other places in the code.


clausnagel:
I found issues with this PR, which require some rework. So I'm putting this PR back into draft mode. @yaozhihang, please postpone your review, thanks.

clausnagel:
Ok, ready for review now. I changed the following:
- The `DatabaseConnectionPool` now takes care of setting the workspace for a connection/session. As a consequence, a lot of code could be removed from several classes of the import, export, and delete operations.
- A connection/session is always put into the default workspace if no workspace is provided by the user. This behaviour is already implemented in the `master` and in previous versions of the Importer/Exporter, but I accidentally removed it in this branch.
- Minor changes to the `Database` tab in the GUI. A user can now request a list of available workspaces (like with schemas).

impexp-client/src/main/java/org/citydb/gui/modules/database/view/DatabasePanel.java
```diff
@@ -591,6 +649,13 @@ private void setDbConnection(DatabaseConnection databaseConnection) {
 		databaseConnection.setUser(userText.getText());
 		databaseConnection.setPassword(String.valueOf(passwordText.getPassword()));
 		databaseConnection.setSavePassword(passwordCheck.isSelected());
+
+		String workspace = getValue(workspaceCombo);
+		if (databaseTypeCombo.getSelectedItem() == DatabaseType.ORACLE && workspace != null) {
+			databaseConnection.setWorkspace(new Workspace(workspace, timestamp.getDate()));
+		} else {
+			databaseConnection.setWorkspace(null);
```
yaozhihang, line 657: This code looks not correct (see the settings in the screenshot). In this case, the workspace assigned to `databaseConnection` should not be NULL, but a `new Workspace("LIVE", timestamp.getDate())`.
image

impexp-core/src/main/java/org/citydb/database/adapter/oracle/WorkspaceManagerAdapter.java
```diff
@@ -74,4 +80,21 @@ public boolean gotoWorkspace(Connection connection, Workspace workspace) {
 		}
 	}
 
+	@Override
+	public List<String> fetchWorkspacesFromDatabase(Connection connection) throws SQLException {
+		try (Statement stmt = connection.createStatement();
+			 ResultSet rs = stmt.executeQuery("select workspace from all_workspaces order by workspace")) {
```
yaozhihang, line 86: Should we only query those workspaces that are not frozen? Maybe also owned by the user (schema)?

yaozhihang (commented):
thanks @clausnagel for the PR. Only 4 comments or questions after my review and tests.

impexp-core/src/main/java/org/citydb/database/connection/DatabaseConnectionPool.java
```diff
@@ -148,42 +153,66 @@ public synchronized void connect(DatabaseConnection connection) throws DatabaseC
...
 					throw new SQLException("The database schema '" + connectionDetails.getSchema() + "' does not exist.");
-			} else
+				}
+			} else {
 				connectionDetails.setSchema(databaseAdapter.getSchemaManager().getDefaultSchema());
-			
+			}
+
+			// check workspace
+			if (databaseAdapter.hasVersioningSupport()) {
```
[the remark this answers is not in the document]
clausnagel: fixed with 9474fe2c5dbb59d26da915c606484e06eb71c6d9

impexp-core/src/main/java/org/citydb/database/connection/DatabaseConnectionPool.java
```diff
@@ -192,22 +221,32 @@ public AbstractDatabaseAdapter getActiveDatabaseAdapter() {
...
 			throw new SQLException("Database is not connected.");
+		}
 
 		Connection connection = dataSource.getConnection();
 		connection.setAutoCommit(true);
 
+		// change workspace if required
+		if (databaseAdapter.hasVersioningSupport()) {
+			Workspace workspace = databaseAdapter.getConnectionDetails().getWorkspace();
+			databaseAdapter.getWorkspaceManager().gotoWorkspace(connection, workspace);
```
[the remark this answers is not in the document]
clausnagel: I added `databaseAdapter.getConnectionDetails().isSetWorkspace()`  to avoid a NPE (see 28e509be11dd54ce92c453eafba050843e6c5149).

Yes, I think it's necessary to call `gotoWorkspace` every time. At least, this is the behaviour in `master` and previous releases. As far as I remember the reason is the use of a connection pool. Assume you connect to the database with workspace `ABC` and execute some operations. Afterwards, you disconnect from the database. Due to the connection pool, the connection is not physically closed. If you now re-connect, for instance, with the default workspace, the connection pool might return the previous connection from `getConnection` that is still on workspace `ABC`. So, you must invoke `gotoWorkspace` to set it to the new workspace correctly.

impexp-client/src/main/java/org/citydb/gui/modules/database/view/DatabasePanel.java
clausnagel: Well, the assumption is that a name is always required if you want to use a timestamp. Does it make sense to set a timestamp on the `LIVE` workspace? If so, then I agree and we must change this behaviour.

impexp-core/src/main/java/org/citydb/database/adapter/oracle/WorkspaceManagerAdapter.java
clausnagel: I am actually not an Oracle expert for this question. Assume that you are user `AAA` but connect to the database using the schema `BBB` instead. And assume that the schema of `BBB` is version-enabled and `BBB` has created some workspaces. Is the user `AAA` allowed to goto these workspaces? I assume the answer is no unless you have been granted sufficient privileges? Can you grant privileges on a workspace?

Regarding frozen workspaces: Well, we have not checked this in previous versions. But you are right, we might have to consider this. What are the criteria to remove a frozen workspace from the list?

impexp-client/src/main/java/org/citydb/gui/modules/database/view/DatabasePanel.java
clausnagel: This should fix it: bc0fb3b264a19d9e41fb93819fd2f131f74f880b. Can you please test again?

yaozhihang: very good

impexp-core/src/main/java/org/citydb/database/adapter/oracle/WorkspaceManagerAdapter.java
clausnagel: Thanks, @yaozhihang, for solving an issue with using workspaces in another schema in the 3DCityDB repo (https://github.com/3dcitydb/3dcitydb/commit/1b00b87f583b6de6153ae61b7748579a22c84fc1).

We kept frozen workspaces in the list of workspaces that can be queried on the Database tab. If the workspace is in `NO_ACCESS` mode, a corresponding error message is now shown to the user (168a5efecf181f52e33bbf3f5dc916f490f22ba1).

yaozhihang (approved):

PR #168, opened 2021-01-29
Add collapsible filter panels

clausnagel:
This PR proposes collapsible filter panels for the `Import`, `Export` and `KML/COLLADA/glTF` tabs. I think this makes the GUI a bit clearer. Let me know your thoughts.


impexp-config/src/main/java/org/citydb/config/gui/exporter/ExportGuiConfig.java
```diff
@@ -0,0 +1,86 @@
...
+        "collapseAttributeFilter",
+        "collapseSQLFilter",
+        "collapseLodFilter",
+        "collapseCounterFilter",
+        "collapseBoundingBoxFilter",
+        "collapseFeatureTypeFilter",
+        "sqlFilter"
+})
+public class ExportGuiConfig {
+    private boolean collapseAttributeFilter = true;
```
yaozhihang, line 17: I'm not sure, if we should disable the collapse option per default for all filter components. My concern is that will such clean GUI astound the users who get used with the old GUI?

yaozhihang (commented):
Very cool implementation. It makes the GUI much more user-friently. Three small comments are left below.

impexp-config/src/main/java/org/citydb/config/gui/exporter/ExportGuiConfig.java
Son-HNguyen: I also think that the most frequently used filters/lists should be shown, and the less frequently used hidden by default. At least the options that are enabled by default should be shown (such as `Tiling` in `KML/COLLADA/glTF Export`). 

Personally in `Import`, I use the section `Bounding Box` a lot. In `Export` the `LoD Filter` and `Bounding Box`. In `KML/COLLADA/glTF Export` the `Tiling` and also `Bounding Box`.

If possible, I would suggest enabling and showing these by default (like in the case of `Tiling` in `KML/COLLADA/glTF Export`).

yaozhihang: @Son-HNguyen thanks for the comment. Sounds reasonable.
@BWibo What's your opinion?

clausnagel: Thanks for the feedback. In general, all filters that are collapsed are optional. Also tiling is not mandatory for `KML/COLLADA/glTF` exports anymore. The mandatory dialogs for `KML/COLLADA/glTF`  are "Export from level of detail" and "Display as" and they cannot be collapsed.

Note that the GUI remembers the collapsed state of all filter dialogs. If you have opened the tiling or bounding box filter for KML/COLLADA/glTF exports, it will also be opened when you start the Importer/Exporter the next time (even if the filter is not activated). So, we are just talking about the first initial start of the GUI.

What I would like to avoid is that one of the tabs starts with a scroll bar. I'm running on a 1920x1080 resolution, and with all filter dialogs opened on the `Export` tab I have no scroll bar on Windows 10 - but there are only a few pixels left. On Ubuntu, I already get a scroll bar. I don't like this because then you do not see all capabilities at one glimpse.

The "most frequently used" is hard to answer because we have no statistics. I agree that for `KML/COLLADA/glTF` exports the tiling filter is still an important one to keep file sizes small - even if it's not mandatory anymore. But for CityGML exports, I rarely use the LoD filter for example.

impexp-config/src/main/java/org/citydb/config/project/exporter/SimpleQuery.java
```diff
@@ -71,8 +75,10 @@
     protected FeatureTypeFilter featureTypeFilter;
     @XmlElement(name = "featureVersion")
     private SimpleFeatureVersionFilter featureVersionFilter;
-    @XmlElement(name = "selection")
-    private SimpleSelectionFilter selectionFilter;
+    @XmlElement(name = "attributes")
+    private SimpleAttributeFilter attributeFilter;
+    @XmlElement(name = "sql", required = true)
```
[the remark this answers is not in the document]
clausnagel: fixed with 827d979454ab123c18037b7bb4b8d2bd5b554a96

impexp-client/src/main/java/org/citydb/gui/modules/exporter/view/FilterPanel.java
```diff
@@ -380,6 +380,14 @@ private void setEnabledLodFilterMode() {
 		lodMode.setEnabled(selected > 1);
 	}
 
+	private void setEnabledAttribtueFilter() {
```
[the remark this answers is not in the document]
clausnagel: fixed with 8c47991c397edc66795eb5c4a0b8cc52f2de76d3

impexp-config/src/main/java/org/citydb/config/gui/exporter/ExportGuiConfig.java
Son-HNguyen: > Note that the GUI remembers the collapsed state of all filter dialogs. If you have opened the tiling or bounding box filter for KML/COLLADA/glTF exports, they will also be open when you start the Importer/Exporter the next time (even if the filter is not activated). So, we are just talking about the first initial start of the GUI.

I also noticed this while testing. This makes it much more comfortable when using the tool in the long run!

clausnagel: Ok, I propose to keep the current behaviour and start with collapsed filter dialogs on the initial first startup of the GUI. We will soon have internal tests at VCS and will get more feedback from users.

I could add a right-click popup menu to each filter label that could offer options to exand/collapse one or all of the filter dialogs. This might be helpful to see all filter parameters with just one click.

yaozhihang: The right-click popup menu is perfect. I think it offers a very good compromise. I agree to keep the initial state of the filter panels to be collpased at the moment.

yaozhihang (approved):

PR #169, opened 2021-02-03
Docker

BWibo:
This PR is migrating the content from https://github.com/tum-gis/3dcitydb-importer-exporter-docker to this repo.

### Todo
- [x] Debian image
- [x] Alpine image
- [x] Github Action for automated build and push on commits to master to ``3dcitydb/impexp:edge``, ``3dcitydb/impexp:edge-alpine``
- [x] Github Action build only for push to pull requests ``3dcitydb/impexp:edge``, ``3dcitydb/impexp:edge-alpine``

### Quickstart
```bash
docker run -i -t --name impexp --rm \
    -u $(id -u):$(id -g) \
    -v /local/share/dir:/share \
  3dcitydb/importer-exporter COMMAND OPTS ARGS
```


BWibo:
I added Dockerfiles and workflows for debian and apline edge images.
Sadly, the ``openjdk`` Docker images do not support alpine for JDK 11, that's why we currently need a different base image for alpine versions and thus a second Dockerfile. In the future, this will hopefully change so we can narrow down to a single Dockerfile.
Docker build checks and build-push workflows ar in place as well. They will work as soon as the secrets are set up.

BWibo:
@yaozhihang @clausnagel 
Do you think it makes sence to build images for other Java versions than JDK/JRE 11 LTS?
 If yes, which versions should we support? 
I could easily setup e.g. build checks for all versions provided [here](https://hub.docker.com/_/openjdk?tab=tags&page=1&ordering=last_updated) (debian) and [here](https://hub.docker.com/r/adoptopenjdk/openjdk11) (alpine).
In my opinion it is fine to release just one image version based on the current Java LTS version to the public.
However, build checks for future Java versions and maybe Java 8 could make sence.

BWibo:
@clausnagel @yaozhihang Hey there, I would like to have this ready for v4.3.0. When do you plan to release?

yaozhihang:
@BWibo Good. So this pull request is _Ready for review_?

BWibo:
Yes, it's ready for review. There is just one thing we should discuss before merge.
According to [Dockerfile best practice](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/), `impexp` should run as non-root user inside the containers.
I added how this would look like here: https://github.com/3dcitydb/importer-exporter/tree/docker-non-root

This is usually something nobody will notice, however, if the user Linux  user and group IDs inside the container are different from the one's on the host system, there might be issues with shared files ownership / groups.
This can easily be avoided by using the `docker run -u` option:

```sh

docker run -i -t \
  -u $UID:$GID \
  -v /my/share/temp:/share 3dcitydb/impexp:nonroot CMD OPTS ARGS
 
```
The UID and GID inside the container is set to 1000, which is the default for the first user create on a Linux system.
Hence, in most cases adding `-u $UID:$GID` will not be required anyway.

If you agree to release non-root image, I'll just merge `docker-non-root` to this before wwe merge this.

One more thing. I need to set the DockerHub credential secrets for the workflows to work (@clausnagel)

clausnagel:
+1 for non-root image

Access rights granted for setting up the workflow.

BWibo:
I rebased the `non-root-image` branch on this one, it is included now. The workflows appear to be working, thx.

BWibo:
@yaozhihang @clausnagel 
Some recent changes currently break the Docker build:
When I merge a0ffc66d887178e41daf7b2434de1ccc44b92d90 (current master) into the `docker` branch, I get following error during build. I believe the issue is the missing `git` dist inside the container.
```
#11 40.99 > Task :impexp-client:update3DCityDBGitSubmodule FAILED
#11 40.99
#11 40.99 FAILURE: Build failed with an exception.
#11 41.00
#11 41.00 * What went wrong:
#11 41.00 Execution failed for task ':impexp-client:update3DCityDBGitSubmodule'.
#11 41.00 > A problem occurred starting process 'command 'git''
```

Will we need this in the future? If yes, I'll add git to the builder image.

yaozhihang:
@BWibo yes, it is an issue. A difficult thing is that we cannot completely get rid of `git`, because the `dockerInstallDist` also requires the submodule of `spreadsheet-generator-plugin`. I think, we need the `.git` folder and `git` and have to endure the longer build time for now.  I'll resolve this issue in future (after release 4.3.0) as soon as we move the current _composite builds_  to the `3dcitydb-suite` repository.

BWibo:
@yaozhihang I added `git` and the `.git` folder is now copied to the builder image. 
However, this only works locally and breaks the Github Actions, as they checkout just one commit. I'll look into this and see if there is a way to change that behavior, even though this not really good. The entire impexp repo would be checked out for every build in that case.

BWibo:
@yaozhihang I did some reading and testing regarding that issue and reverted (force push, not revert, sorry) the above mentioned changes. It is possible to update submodules recursively using Github actions. I  am wondering if it would be possible to have an `dockerInstallDisk` task, that does not use the Gradle `updateSubModule` routines?
Maybe we can discuss options via zoom in the upcoming days.

clausnagel:
@BWibo, @yaozhihang, I have merged the latest from master into this branch and slightly adapted the docker files so that they work again. Please review and comment.

@BWibo, when running CLI commands using the Docker container, I found it quite useless that with every command a banner with maintainer information and the passed CLI options are pinted to the console. The passed CLI options are recorded in the log file, so no need to copy them to the console. And we can put the maintainer information, for instance, in the user manual, a readme file and a description on docker hub. But I would remove them from the entrypoint script.

If we agree on removing the banner and the repeated CLI options, then I think there is no need for the entrypoint script anymore?

BWibo:
Yes, we don't need the entrypoint script anymore. I removed it and add ``help`` as default CMD if no command is appended:  `docker run 3dcitydb/impexp <COMMAND>`.
I did some quick tests and everything seems fine with the new images. I pushed the versions I built as 
`3dcitydb/impexp:edge` and `3dcitydb/impexp:edge-alpine` so you can try them out too.

clausnagel:
Thanks, @BWibo. You were faster than me with your commits :-) I have also made some smaller changes to the Dockerfiles. I did not commit them but instead created a patch file based on your last commits in this branch: [docker.zip](https://github.com/3dcitydb/importer-exporter/files/7038269/docker.zip).

My main intention was to tidy up the scripts a bit. And I added the home folder of the user as default WORKDIR. This way you can import/export files by just providing a filename (not an entire path). And I re-added an entrypoint script to allow the image to be used in OpenShift environments, where an arbitrary user might be created on container start. Don't know how important that is but found it in other Dockerfiles as well.

BWibo:
Oh sorry for the confusion.
I checked the patch but I can't find the new `entrypoint.sh` script. Am I missing something?

clausnagel:
Yes, new files are of course not part of a patch file, sorry... Download from here as ZIP file [impexp-entrypoint.zip](https://github.com/3dcitydb/importer-exporter/files/7038644/impexp-entrypoint.zip) and put it into resources/docker

clausnagel:
If you agree with the above proposal to change the Docker files, I can create a commit for it

BWibo:
Hey, thx. Looks good. I did some testing and it seems fine.
I'll have to do some more reading regarding the OpenShift stuff. We might need to add changing folder permissions of the `$HOME` folder in the container to make that work for arbitrary users.
I hope I find the time to try that out tomorrow.

BWibo:
I pushed the newly created images for testing:
* `docker pull 3dcitydb/impexp:edge`
* `docker pull 3dcitydb/impexp:edge-alpine`

BWibo:
Yes, if this is merged to `master`, the `edge` and `edge-alpine` tags should be built and pushed to Dockerhub.
Same applies for the other images, when a new release is created after the merge. I'll test that in a fork tomorrow.
I just removed the build args from the workflows, so we don't have to update base image versions in multiple places. That means that the workflows always use the default we provide in the `Dockerfiles`.

BWibo:
I create a fork to test the workflows and all actions seem to work fine.
The [actions]( https://github.com/BWibo/importer-exporter/actions) passed all and the images have been pushed.
I change the repo to [3dcitydb/impexp-test](https://hub.docker.com/repository/docker/3dcitydb/impexp-test) for the tests. Take a look at the DockerHub page to see the result. When everything is fine, I'll delete the repo from DockerHub.
With ~ 100MB size we have some really lightweight images now!

I added a few minor adjustments:

* `build-push-release` workflows: added creation and push of `latest` tag
* `build-edge`: workflows: This is now only triggered for pull request, as on push to master `build-push-edge` runs
* `all workflows`: Automatic use of latest minor release of `docker-build-push` action

I think this is ready for merge now.

BWibo:
I just realised that I accidently merged stuff from test test repo to `master` of this repo. I'm sorry. 
I force pushed `master` back to ist old state, but I can't reopen this PR. Can you @clausnagel?

clausnagel:
No, it seems that you cannot reopen pull requests once they are merged :-(

BWibo:
I found this: https://gist.github.com/robertpainsi/2c42c15f1ce6dab03a0675348edd4e2c

clausnagel:
hm, I think the merge didn't work - the version of the files in master is not the latest version we discussed here

BWibo:
We could open a new PR and reference this one. Sorry for causing this mess...

PR #172, opened 2021-02-16
Reworked delete list processing

clausnagel:
The delete list is now loaded into a temporary database table. This allows us to
- join the temporary table with ` CITYOBJECT` to get a precise number of affected objects in preview mode
- easily combine a delete list with other filter settings in a single database query. The CLI delete command has been adapted correspondingly.


yaozhihang (approved):
@clausnagel Thanks for the MR. Very good!

PR #175, opened 2021-02-22
Add CLI options for feature version filter

clausnagel:
This PR adds CLI options to the `export`, `export-vis` and `delete` commands that allow to control the feature version filter.

One breaking change: 
* The glTF export option `--remove-collada` had a shortcut option  `-r` before. Since `-r` is now used by the feature version filter, the new shortcurt option for removing collada files is `-m`.

### DIFF
--- a/impexp-plugin-api/src/main/java/org/citydb/plugin/cli/FeatureVersionOption.java
+++ b/impexp-plugin-api/src/main/java/org/citydb/plugin/cli/FeatureVersionOption.java
@@ -1,0 +1,155 @@
+package org.citydb.plugin.cli;
+
+import org.citydb.config.project.query.simple.SimpleFeatureVersionFilter;
+import org.citydb.config.project.query.simple.SimpleFeatureVersionFilterMode;
+import picocli.CommandLine;
+
+import javax.xml.datatype.DatatypeConstants;
+import javax.xml.datatype.DatatypeFactory;
+import javax.xml.datatype.XMLGregorianCalendar;
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.chrono.IsoChronology;
+import java.time.format.DateTimeFormatter;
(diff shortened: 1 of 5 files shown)


impexp-plugin-api/src/main/java/org/citydb/plugin/cli/FeatureVersionOption.java
```diff
@@ -0,0 +1,155 @@
...
+import java.time.format.DateTimeParseException;
+import java.time.format.ResolverStyle;
+import java.time.temporal.ChronoField;
+
+public class FeatureVersionOption implements CliOption {
+    @CommandLine.Option(names = {"-r", "--feature-version"}, required = true,
+            description = "Feature version: ${COMPLETION-CANDIDATES}.")
+    private Version version;
+
+    @CommandLine.Option(names = {"-R", "--feature-version-timestamp"}, paramLabel = "<timestamp[,timestamp]>",
```
yaozhihang, line 25: The timestmap can also be given as a UTC datetime with Zone Offset like `2021-02-22T05:04:03+02:00`. Maybe this should be mentioned in the description.

yaozhihang (commented):
Thanks @clausnagel. The MR looks good to me. I have only one small remark (see my comment below).

yaozhihang:
@BWibo This MR changes a CLI shortcut option of glTF export. You may need to check if it affects your docker-based test environment.

impexp-plugin-api/src/main/java/org/citydb/plugin/cli/FeatureVersionOption.java
clausnagel: You are right, it's supported. I adapted the format string for date-time in the description to `YYYY-MM-DDThh:mm:ss[(+|-)hh:mm]`. I wanted to use the plusminus `±` character instead of `(+|-)` but it's not rendered correctly in all terminals.

yaozhihang (approved):

PR #176, opened 2021-02-24
added support for operating database bounding box with feature version filter

yaozhihang:


clausnagel:
Thanks @yaozhihang. As discussed offline, I moved the logic of the `createBoundingBoxes` method from the database adapter to the operation panel and added checks whether the feature version filter settings are correct.

yaozhihang:
Thanks @clausnagel. The PR looks better and also works well with my tests.

PR #179, opened 2021-04-15
fixed join of unregistered ADE hook tables

yaozhihang:


clausnagel:
Thanks for spotting this bug, @yaozhihang. I checked your fix and would like to propose to rather fix the `getADEHookTables()` method of `CityGMLExportManager`. This way, the class `AbstractTypeExporter` can just rely on the returned result when calling the method. Otherwise every caller of the method would have to validate the result.

[ade-export-fix.zip](https://github.com/3dcitydb/importer-exporter/files/6325353/ade-export-fix.zip)

yaozhihang:
Thanks Claus, your patch make the fix better. Included in 594df83

clausnagel (approved):

PR #180, opened 2021-04-16
Add metadata options to the import and delete CLI commands

clausnagel:
This PR adds options to set the lineage, updating person, and reason for update for the `import` and `delete` CLI commands. For `import`, additional options have been added to set the creation date mode and termination date mode.


yaozhihang:
Thanks @clausnagel. Nice enhancement of the CLI interface. I have only one remark: The `importCityDBMetadata` is missing the MetadataOption of the import operation. Would be good to also include it.

clausnagel:
I added a `--use-metadata-from-file` flag. If set, the values for lineage, updating person and reason for update from the input file take precedence over the values provided through the `--lineage`, `--updating-person`, and `--reason-for-update` options. Note that this does not affect creationDate and terminationDate, because they have their own individual modes.

yaozhihang (approved):

PR #181, opened 2021-04-17
Add possibility to create delete log

clausnagel:
This PR adds the `--delete-log` option to the delete CLI command, which allows for creating a log CSV file listing the deleted/terminated top-level features.

This PR is based on #180.

### DIFF
--- a/impexp-core/src/main/java/org/citydb/citygml/deleter/controller/Deleter.java
+++ b/impexp-core/src/main/java/org/citydb/citygml/deleter/controller/Deleter.java
@@ -110,7 +110,7 @@ public class Deleter implements EventHandler {
 			try {
 				Path logFile = config.getDeleteConfig().getDeleteLog().isSetLogFile() ?
 						Paths.get(config.getDeleteConfig().getDeleteLog().getLogFile()) :
-						CoreConstants.IMPEXP_DATA_DIR.resolve(CoreConstants.IMPORT_LOG_DIR);
+						CoreConstants.IMPEXP_DATA_DIR.resolve(CoreConstants.DELETE_LOG_DIR);
 				deleteLogger = new DeleteLogger(logFile,
 						config.getDeleteConfig().getMode(),
 						config.getDatabaseConfig().getActiveConnection());
--- a/impexp-core/src/main/java/org/citydb/citygml/deleter/controller/Deleter.java
+++ b/impexp-core/src/main/java/org/citydb/citygml/deleter/controller/Deleter.java
@@ -33,12 +33,13 @@ import org.citydb.citygml.common.cache.model.CacheTableModel;
 import org.citydb.citygml.deleter.DeleteException;
 import org.citydb.citygml.deleter.concurrent.DBDeleteWorkerFactory;
 import org.citydb.citygml.deleter.database.BundledConnection;
+import org.citydb.citygml.deleter.database.DBSplittingResult;
 import org.citydb.citygml.deleter.database.DeleteListImporter;
 import org.citydb.citygml.deleter.database.DeleteManager;
 import org.citydb.citygml.deleter.util.DeleteListException;
 import org.citydb.citygml.deleter.util.DeleteListParser;
+import org.citydb.citygml.deleter.util.DeleteLogger;
 import org.citydb.citygml.deleter.util.InternalConfig;
-import org.citydb.citygml.exporter.database.content.DBSplittingResult;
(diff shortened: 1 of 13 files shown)


impexp-core/src/main/java/org/citydb/citygml/deleter/controller/Deleter.java
```diff
@@ -101,15 +104,39 @@ public boolean doDelete(boolean preview) throws DeleteException {
 		eventDispatcher.addEventHandler(EventType.OBJECT_COUNTER, this);
 		eventDispatcher.addEventHandler(EventType.INTERRUPT, this);
 
+		// create delete logger
+		DeleteLogger deleteLogger = null;
+		if (!preview && config.getDeleteConfig().getDeleteLog().isSetLogDeletedFeatures()) {
+			try {
+				Path logFile = config.getDeleteConfig().getDeleteLog().isSetLogFile() ?
+						Paths.get(config.getDeleteConfig().getDeleteLog().getLogFile()) :
+						CoreConstants.IMPEXP_DATA_DIR.resolve(CoreConstants.IMPORT_LOG_DIR);
```
yaozhihang, line 113: use `CoreConstants.DELETE_LOG_DIR`

impexp-core/src/main/java/org/citydb/citygml/deleter/database/DeleteManager.java
```diff
@@ -261,7 +266,23 @@ private void doDelete(Select select, long hits) throws SQLException {
...
+		if (deleteLogger != null) {
+			try (PreparedStatement stmt = databaseAdapter.getSQLAdapter().prepareStatement(select, connection);
+				 ResultSet rs = stmt.executeQuery()) {
+				while (rs.next()) {
+					long id = rs.getLong("id");
+					int objectClassId = rs.getInt("objectclass_id");
+					String gmlId = rs.getString("gmlid");
+					AbstractObjectType<?> objectType = schemaMapping.getAbstractObjectType(objectClassId);
+
+					deleteLogger.write(objectType != null ? objectType.getPath() : String.valueOf(objectClassId), id, gmlId);
```
yaozhihang, line 279: Hm, i think we should not allow terminating NULL-objectType (e.g. ADE classes which are registered in DB, but not supported by impexp), in order to keep consistent behaviour in delete and export processes.

yaozhihang:
@clausnagel Thanks for the nice PR. Please check my two comments of my review

impexp-core/src/main/java/org/citydb/citygml/deleter/controller/Deleter.java
clausnagel: fixed with dfc1d9d619cc3540360d7d1341f6598549f9db15

impexp-core/src/main/java/org/citydb/citygml/deleter/database/DeleteManager.java
clausnagel: Terminate fails now in case the `objectType` cannot be derived is thus is null (see fefd0cf32f24162d0f2a95828cf74125bb90f6db).

yaozhihang: Looks OK to me for this PR. But this would work only when the delete logger is enabled. I think we could do some furhter tweaks in a separate PR.

yaozhihang (approved):

impexp-core/src/main/java/org/citydb/citygml/deleter/database/DeleteManager.java
clausnagel: Yes, we should definitely put this discussion into a separate PR because this PR is just about adding a delete log.

Well, the reason for having this restriction for the import and export operation is simply because we need additional code to be able to parse and write ADE datasets. When deleting, we do not need to parse/write ADE datasets. So I am not convinced that we really should restrict the delete operation in this way.

clausnagel: Ok, one more note on this: I think the delete operation already behaves like you expect (even before this PR). I tested by registering the TestADE and importing a TestADE dataset. Afterwards, I removed the TestADE extension from the Importer/Exporter (so that the TestADE is still registered in DB but not supported by the Importer/Exporter).

Afterwards, I executed a delete operation. TestADE features are not deleted/terminated in this case.

Ok, will merge this PR now.

PR #183, opened 2021-04-21
use JFrog Artifactory Cloud for deploy and publish

yaozhihang:


clausnagel:
I moved all other 3DCityDB artifacts to citydb.jfrog.io as well. There is a similar `jfrog-artifactory-cloud` branch in all corresponding GitHub repos. And the Oracle libraries are now also loaded dynamically from the Maven repository and not stored as static files in the repo anymore.

PR #184, opened 2021-04-22
Fix export of ADE properties for ADE features directly derived from AbstractCityObject

clausnagel:


impexp-core/src/main/java/org/citydb/citygml/exporter/database/content/DBCityObject.java
```diff
@@ -472,10 +482,79 @@ private void postprocess() throws CityGMLExportException, SQLException {
 		generalizesToExporter.executeBatch();
 	}
 
+	private void delegateADEProperties(ObjectContext context, long objectId) throws SQLException, CityGMLExportException {
+		// we only have to query and delegate ADE properties here if the
+		// ADE feature is a direct child of AbstractCityObject and, thus,
```
yaozhihang, line 487: better to say here "direct child of AbstractCityObject or AbstractSite"

yaozhihang (approved):
Looks good.

impexp-core/src/main/java/org/citydb/citygml/exporter/database/content/DBCityObject.java
clausnagel: good point, implemented with 1b812ecd32b23d66efdfd99e17026bc77d0a3f6a

PR #189, opened 2021-05-13
Make StatementFinalizer thread-safe or remove it

clausnagel:
The `DatabaseConnectionPool` uses the `StatementFinalizer` interceptor of the Tomcat JDBC Connection Pool, which is used to keep track of all created statements  and to close these statements when the connection is returned to the pool. The implementation of `StatementFinalizer` is not thread-safe and assumes that all statements for a connection are created within the same thread.

The Importer/Exporter generally observes this rule with one exception: temporary tables of the internal cache are populated during imports by `DBImportXlinkWorker` but are queried from `DBImportXlinkResolverWorker`. Thus, statements are created from different threads. This leads to an NPE in `StatementFinalizer` when closing the connection of a temporary table and, even worse, the connection cannot be released from the pool.   

This PR adds the class `ConcurrentStatementFinalizer`, which is a thread-safe version of `StatementFinalizer` and resolves the above issue. The issue could alternatively be solved by not using `StatementFinalizer` at all. It is only an optional utility class that makes sure that all statements are closed to avoid resource leaks. However, I personally like this automatic pool maintenance and would therefore not drop  `StatementFinalizer`.

### DIFF
--- a/impexp-core/src/main/java/org/citydb/citygml/common/cache/CacheTableManager.java
+++ b/impexp-core/src/main/java/org/citydb/citygml/common/cache/CacheTableManager.java
@@ -168,7 +168,7 @@ public class CacheTableManager {
 				cacheConnection.rollback();
 				cacheConnection.close();
 			} catch (SQLException e) {
-				//
+				log.error("Failed to close temporary cache connection.", e);
 			}
 
 			if (databaseConnection != null) {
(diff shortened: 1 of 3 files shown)


yaozhihang:
Nice Fix! Only one question: should the other methods `closeInvoked`, `setProperties`, and `reset` also be _synchronized_  for thread-safe?

clausnagel:
Good point, @yaozhihang. According to my tests and understanding, `closeInvoked`, `setProperties`, and `reset` are only invoked when creating or closing a connection. Since we always do this in a single thread and only the creation of statements happens in multiple threads, I only added `synchronized` to the `createStatement` method. In my tests this was sufficient.

yaozhihang (approved):

PR #194, opened 2021-06-27
Add plugins overview

clausnagel:
This PR adds a "Plugins" section to the preferences tab of the Importer/Exporter GUI, which lists all installed and available plugins. Each plugin is displayed with additional information such as its version, its vendor, a description and the implemented extensions of the Plugin API. In addition, the plugins can be enabled/disabled using this dialog.

The Plugin API has been changed in a non-backwards compatible way to make this new dialog possible. Every plugin must now provide a `plugin.xml` file that is stored in the same package as the plugin class itself and that provides the additional metadata. In addition, an optional `pluginLogo.svg` (and a `pluginLogo_dark.svg` for the dark theme) can be provided.

The new "Plugins" section can be tested with the [ADE Manager plugin](https://github.com/3dcitydb/plugin-ade-manager) and the [Spreadsheet Generator plugin](https://github.com/3dcitydb/plugin-spreadsheet-generator). Simply use the `feature-new-plugin-api` branch for both plugins.


yaozhihang:
Thanks @clausnagel. Very cool new feature.

I have two remarks after my review:

1. How about adding an additional CLI option like `--enabledPlugins`? This could be useful when users want to enable/disable certain `FeatureExportExtensions`
2. It might be a _Nice to Have_ to allow users to change the order of plugins listed in the preference panel. A benefit is that the plugins of `FeatureExportExtensions` can be run by following an user-defined order to build a process chain. 

All other things work well in during my tests. 👍

clausnagel:
Thanks, @yaozhihang.

re 1: Should `--enabledPlugins` only list all enabled/disabled plugins, or would you like to be able to actively enable/disable a plugin on the command line? In the latter case, what would be the input to the paramter? The class name of the plugin?

re 2: In general, you cannot define dependencies for plugins (for instance, you cannot say that plugin B should only be executed after plugin A). The plugins overview simply lists the plugins in alphabetical order.

I agree that it would be nice to be able to define the execution order for separate `FeatureExportExtensions`. For other extensions like `ConfigExtension` or `ViewExtension`, an execution order does not make sense though. We could add a `<depends-on>` element to the `plugin.xml` configuration to express a dependency - this would be static then (and again would not make sense for all plugins). But what if the plugin referenced by `<depends-on>` is not installed or disabled? If we want to have it dynamic, I would rather not use the general plugins overview for this purpose but would recommend a specific dialog. This dialog should only list those plugins for which defining an execution order makes sense (could be a separate "Execution order" node beneath the "Plugins" node in the preferences tree). Hm, sounds doable but requires more effort.

yaozhihang:
1. The `--enabledPlugins` should allow to actively enable a list of plugins. I think we could use the full class name of plugin e.g. `org.citydb.plugins.spreadsheet_gen.SPSHGPlugin`
2. I prefer your latter idea. Yes, you are right. it sounds doable and need more coding. We could move this to a separate PR.

clausnagel:
@yaozhihang, I added a `--use-plugin` option that allows to enable or disable plugins on the command line.

I agree to move your second proposal to another PR once we have a requirement to support the execution order of specific plugins.

yaozhihang:
@clausnagel  Looks very good.

yaozhihang (approved):

PR #195, opened 2021-07-01
Feature import plugin

clausnagel:
This PR adds a feature import extension for plugins similar to the export plugin.

### DIFF
--- a/impexp-core/src/main/java/org/citydb/core/operation/importer/concurrent/DBImportWorker.java
+++ b/impexp-core/src/main/java/org/citydb/core/operation/importer/concurrent/DBImportWorker.java
@@ -215,7 +215,7 @@ public class DBImportWorker extends Worker<CityGML> implements EventHandler {
 				}
 
 				long id;
-				if (work instanceof Appearance) {
+				if (feature instanceof Appearance) {
 					// global appearance
 					id = importer.importGlobalAppearance((Appearance) feature);
 					if (id != 0) {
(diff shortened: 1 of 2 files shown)


impexp-core/src/main/java/org/citydb/core/operation/importer/concurrent/DBImportWorker.java
```diff
@@ -190,56 +193,71 @@ private void doWork(CityGML work) {
...
+						} catch (PluginException e) {
+							throw new CityGMLImportException("Import plugin " + plugin.getClass().getName() + " threw an exception.", e);
+						}
+					}
+				}
 
-			else if (work instanceof AbstractFeature) {
-				AbstractFeature feature = (AbstractFeature)work;
+				long id;
+				if (work instanceof Appearance) {
```
yaozhihang, line 218: I think we should use the variable `feature` instead of `work`, because the `feature` reflects the latest state for the CityGML import

yaozhihang (commented):
Thanks @clausnagel for the extension. I have only one comment.

impexp-core/src/main/java/org/citydb/core/operation/importer/concurrent/DBImportWorker.java
clausnagel: Yes, changed with 1b4e514

yaozhihang (approved):

impexp-core/src/main/java/org/citydb/core/operation/importer/concurrent/DBImportWorker.java
yaozhihang: Nice

PR #196, opened 2021-07-02
Add affine transformations to export operation

clausnagel:
This PR adds affine transformations to the export operation. The implementation mirrors the code of the import operation.


yaozhihang (approved):
@clausnagel Thanks for the PR. Everything looks good to me.

PR #199, opened 2021-07-09
Refactoring of cache table manager

clausnagel:
This PR proses some refactoring of the cache table manager:

1. As before, the cache is initialized with a primary mode (database or local) according to user settings. But sometimes it is required to cache data either in the database or locally independent of this primary mode. The cache manager can handle this situation much more flexible now.
2. One cache is now reused for an entire import, export or delete session (instead of re-creating the cache for every file). This allows to use the cache for data that must be available throughout the entire session.


impexp-core/src/main/java/org/citydb/core/operation/common/cache/CacheTableManager.java
```diff
@@ -27,241 +27,252 @@
...
 
 	public CacheTable createCacheTable(CacheTableModel model) throws SQLException {
-		return createCacheTable(model, cacheConnection, cacheAdapter);		
+		return getOrCreateCacheTable(primaryCache, model, false, CacheTable.class);
 	}
 
-	public CacheTable createCacheTableInDatabase(CacheTableModel model) throws SQLException {
-		initDatabaseConnection();
-		return createCacheTable(model, databaseConnection, databaseAdapter);
+	public CacheTable createCacheTable(CacheTableModel model, CacheMode mode) throws SQLException {
```
yaozhihang, line 74: If we use the API like this:

```
CacheTable table1 = cacheTableManager.createCacheTable(CacheTableModel.ID_LIST, CacheMode.DATABASE);
CacheTable table2 = cacheTableManager.createCacheTable(CacheTableModel.ID_LIST, CacheMode.LOCAL);

```

In this case, the `table1` and `table2` are identical. It would be better that the second code line should throw an exception like "_cacheTableMode.ID_LIST shall not be created twice, even with different CacheMode_...."

yaozhihang (commented):
Thanks @clausnagel for the refactorying. Have three comments.

impexp-core/src/main/java/org/citydb/core/operation/deleter/controller/Deleter.java
```diff
@@ -150,6 +136,22 @@ private boolean process(boolean preview, DeleteLogger deleteLogger) throws Delet
 			}
 		}
 
+		// create delete logger
+		deleteLogger = null;
```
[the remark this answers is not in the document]
clausnagel: fixed with 3d56ad34d8c5433a2387baa97b50911bb00ba0f4

impexp-core/src/main/java/org/citydb/core/operation/common/cache/CacheTableManager.java
```diff
@@ -27,241 +27,252 @@
...
 
-		} finally  {
-			// clean up
-			cacheTables.clear();
-			branchCacheTables.clear();
+		for (BranchCacheTable branchCacheTable : branchCacheTables.values()) {
+			branchCacheTable.drop();
+		}
+
+		for (Cache cache : caches.values()) {
```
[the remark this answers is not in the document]
clausnagel: I would prefer keeping it in the `dropAll` method to not just drop all the tables but to also undoe all additional changes and to release internal resources. Just to make sure the cache is really clean afterwards.

impexp-core/src/main/java/org/citydb/core/operation/common/cache/CacheTableManager.java
clausnagel: The behaviour is intended to be this way. Assume the `createCacheTable` method is invoked by two concurrent threads for the same CacheTableModel. Then both threads will receive a valid reference to the same table and write data to it. Otherwise, the creation of cache tables would always have to happen in a synchonized way.

clausnagel: Good point. An exception is thrown now as suggested. 4b60ab9a3a3d0900c5dcbc4f4422eece601819a5

yaozhihang (approved):

PR #202, opened 2021-07-12
Add import ID list filter

clausnagel:
This PR proposes to add an import list filter to the import operation for both the GUI and the CLI. An import list is a CSV file containing one feature identifier (gml:id) per line (plus possibly additional columns). The identifiers from the import list are checked against the identifiers of top-level features from the input file(s). The user can decide whether a) only features from the list shall be imported or b) all features from the list shall be skipped.

One immediate use case for this new filter option is when using _import logs_ for the import operation. Assume you start an import operation on a set of input files and the import fails after a certain amount of features. You can now **resume the import** after having fixed the issues that caused the failure. Simply use the import log as filter and skip the features that have already been imported according to the import log file.

This PR depends on #199 

image


yaozhihang (approved):
@clausnagel Great enhancement for CityGML Import! All looks good to me and works well in my tests.

PR #207, opened 2021-08-24
Use bigint as data type for ID columns

clausnagel:
This PR implements the changes that are required to support https://github.com/3dcitydb/3dcitydb/pull/66


clausnagel:
Putting this PR back in draft mode. The delete worker works fine with the latest database schema using bigints, but fails to correctly deal with older schemas. Will fix it.

clausnagel:
ready for review again

yaozhihang (commented):
https://github.com/3dcitydb/importer-exporter/blob/feature-use-bigint/impexp-core/src/main/java/org/citydb/core/operation/common/cache/model/CacheTableObjectGmlId.java#L62

here should use `getBigint()`

yaozhihang:
https://github.com/3dcitydb/importer-exporter/blob/feature-use-bigint/impexp-core/src/main/java/org/citydb/core/database/adapter/postgis/UtilAdapter.java#L193
and 
https://github.com/3dcitydb/importer-exporter/blob/feature-use-bigint/impexp-core/src/main/java/org/citydb/core/database/adapter/oracle/UtilAdapter.java#L198
should use `Type.BIGINT`

clausnagel:
Thanks, fixed both comments.

Interestingly, the `createBoundingBox` also works fine with the new `env_cityobject` function when using `Types.INTEGER`. But this doesn't feel right, so `Type.BIGINT` is used instead now if the 3DCityDB version is greater or equal to 4.2.0.

yaozhihang (approved):
Well done @clausnagel

PR #212, opened 2021-08-28
Add environment variables for database connections

clausnagel:
This PR adds support for the following environment variables: `CITYDB_TYPE`, `CITYDB_HOST`, `CITYDB_PORT`, `CITYDB_NAME`, `CITYDB_SCHEMA`, `CITYDB_USERNAME`, and `CITYDB_PASSWORD`. 

When running the Importer/Exporter on the command line, the values of these variables will be used as input if a corresponding CLI option *is not available*. Thus, the CLI options always take precedence. In case the Importer/Exporter is started in GUI mode, the environment variables are ignored.

With the new environment variables, you can, for example, define all database connection details except the password as options on the command line and provide the password using the `CITYDB_PASSWORD` variable.


yaozhihang (approved):
Thanks @clausnagel, very nice feature.

PR #216, opened 2021-09-24
Version info

BWibo:
This introduces the env var IMPEXP_VERSION that is set by github workflows:
* `edge`,  `edge-alpine`: Short SHA of the commit used for teh image
* Release version and `latest`: Release version, e.g. `4.2.0`

Second, the major version of github actions is now used, to received fixes if available:
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsuses


clausnagel:
Hi @BWibo, this PR is marked to be `work-in-progress`. Are you still working on it, or is the PR ready for review?

BWibo:
It has been a while, I'll check that later this week.

yaozhihang:
@BWibo interesting PR. Where the environment variable `IMPEXP_VERSION` will/should be used?

BWibo:
This gives an easy way to retieve the ImpExp version, that used in a Docker image. We can use it e.g. for logging or testing. This is especially useful for the `edge` version, because you can see the exact version (commit sha) the is used in the image.

BWibo:
I see, maintaining this causes some work we should avoid. However, we don't really need to maintain this, as the value is alway overwritten with the correct value in the build workflows, e.g. here:
https://github.com/3dcitydb/importer-exporter/blob/056f538201fd5c0c2f514aa73486f08e26b1d2b6/.github/workflows/docker-build-edge-alpine.yml#L26-L28
I would like to keep the build arg to be able to set the version number during build, but we could simply leave it empty or set it to a value indicating, that the version has not been set.
According to docs it is not required to provide a default value we would need to maintain.
What do you think?

BWibo:
What I proposed aboved results in this, when leaving the build arg unset during build:

````console
$ docker inspect citydb-image
...
PATH=/usr/local/openjdk-11/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
JAVA_HOME=/usr/local/openjdk-11
LANG=C.UTF-8
JAVA_VERSION=11.0.12
IMPEXP_VERSION=
...
```

clausnagel:
What about improving `impexp --version` instead? I guess it can be made more generic so that the version to be returned can be set from the Gradle build script. And maybe we can even make sure that the SHA of the last commit is added in case the build is triggered from `master`?

BWibo:
That would be great anyway. Both the CLI and GUI version could benefit from that. I did something similar for another Java/Gradle project. I added the code I used below. The solution calls ``git`` which needs to be present in the build env. Possibly, there are better solutions but maybe this is helpful or can be partially reused.

If ``impexp --version`` provides the extra details, I think we could drop ``IMPEXP_VERSION`` build arg and env variable. There is one good thing about having the variable. Calling ``docker inspect IMAGE`` gives information about the version of the Importer/Exporter inside the image, without having to run anything, but that is not too important.

```groovy
// Jar config ---------------------------------------------------------------
    jar {
        manifest {
            attributes('Implementation-Title': project.name,
                    'Implementation-Version': project.version,
                    'Implementation-Vendor': project.vendorName + ', ' + project.vendorOrganisation
            )
        }
    }

    task versionTxt() {
        doLast {
            sourceSets.main.output.resourcesDir.mkdirs()
            new File(sourceSets.main.output.resourcesDir, "version.txt").text = """Project: ${project.name}
Organization: ${project.vendorName},  ${project.vendorOrganisation}
Author: ${project.author} [${project.authorEmail}]
-- Build info ---------------------------------
Build time: ${getBuildDateTime}
Commit time: ${commitTime}
Branch: ${branchName}
Revision: ${commitHash}
Latest tag: ${tag}
"""
        }
    }

// Functions ------------------------------------------------------------------
import java.text.SimpleDateFormat

def getBuildDateTime = { ->
    def date = new Date()
    SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss")
    // println sdf.format(date)
    return sdf.format(date)
}

def branchName = { ->
    def branch = new ByteArrayOutputStream()
    exec {
        commandLine 'git', 'rev-parse', '--abbrev-ref', 'HEAD'
        standardOutput = branch
    }
    return branch.toString().trim()
}

def commitHash = { ->
    def hash = new ByteArrayOutputStream()
    exec {
        commandLine 'git', 'rev-parse', '--short', '--verify', 'HEAD'
        standardOutput = hash
    }
    return hash.toString().trim()
}

def tag = { ->
    def tag = new ByteArrayOutputStream()
    exec {
        commandLine 'git', 'describe', '--tags', '--abbrev=0'
        standardOutput = tag
    }
    return tag.toString().trim()
}

def commitTime = { ->
    def committime = new ByteArrayOutputStream()
    exec {
        commandLine 'git', 'show', '-s', '--format=%cd', '--date=format:%Y-%m-%d_%H-%M-%S', commitHash()
        standardOutput = committime
    }
    return committime.toString().trim()
}

// Gets the version name from the latest Git tag
def getVersionName = { ->
    getBuildDateTime()
    return commitTime() + '-' + tag() + '-' + commitHash()
}
```

clausnagel:
Ok, so for now, we can merge this PR into `master` and continue the discussion outside this PR.

Issue #223, opened 2021-11-03
Import log file is not created correctly when using non-default path


yaozhihang:
How to reproduce the issue:
1. enable the import-log option in preferences settings and choose a custom path e.g. "D:\imported-features"
2. run CityGML Import. 
3. If the "imported-features" folder doesn't exist, a log file with the name "imported-features" will be created. This looks not correct. The expected behaviour is that a new folder "imported-features" should be created and there a log file with the name somehow like "imported-features-2021-11-03_10-07-01-007" should be created.

clausnagel:
The current implementation is indeed different to what you expect. You can either provide a folder _or_ a file. The idea is that you can also pick your own target file if you do not want to use `imported-features-<timestamp>.log`.

For instance, if you provide `/my/own/log` as target path, then the following steps are performed:

1. If `log` is a folder and if it already exists, then the importer will create the log file in this folder and use `imported-features-<timestamp>.log` as filename.
2. If `log` does not exist, the importer assumes that this shall be the filename of the log file. It will create all intermediate folders (like `my` and `own` in the above example) if they do not exist. And the log is written to a file named `log`.

That explains why in your example all log entries are written to `imported-features`. When checking the user manual, it seems that this behavior is not well described though.

PR #228, opened 2021-11-12
Change import log file behavior

clausnagel:
* The user must now provide a path to a file. Providing a directory is not supported anymore.
* Added the three log file modes `append`, `truncate` and `unique`
  * With `append`, multiple import operations can be logged to the same file
  * When choosing `truncate`, the target log file is truncated before every import
  * `unique` creates a unique file name by appending a timestamp as suffix. This mode reflects the previous behavior.

This PR also addresses #223.


yaozhihang:
@clausnagel thanks. Looks good. Only one remark: The two new checkboxes `truncateLogFile` and `uniqueFileName` should be disabled once the import log has been disabled. 

```
		truncateLogFile.setEnabled(logFeatures.isSelected());
		uniqueFileName.setEnabled(logFeatures.isSelected());
```

clausnagel:
fixed with bf16bf80570d772ed40e200448d33b38be4b9df3

PR #235, opened 2022-01-07
Add option to control whether to compute the number of matching top-level objects

clausnagel:
On large databases, computing the number of top-level objects matching the provided query can take a long time. This PR adds an option to the general export preferences, which allows a user to disable this feature. If disabled, the number of matching top-level objects is no longer recorded in the log and the GUI cannot use a progress bar to indicate the export progress.

### DIFF
--- a/impexp-config/src/main/java/org/citydb/config/project/exporter/ComputeNumberMatched.java
+++ b/impexp-config/src/main/java/org/citydb/config/project/exporter/ComputeNumberMatched.java
@@ -1,0 +1,57 @@
+/*
+ * 3D City Database - The Open Source CityGML Database
+ * https://www.3dcitydb.org/
+ *
+ * Copyright 2013 - 2022
+ * Chair of Geoinformatics
+ * Technical University of Munich, Germany
+ * https://www.lrg.tum.de/gis/
+ *
+ * The 3D City Database is jointly developed with the following
+ * cooperation partners:
+ *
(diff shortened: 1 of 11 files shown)


impexp-config/src/main/java/org/citydb/config/project/exporter/ComputeNumberMatched.java
```diff
@@ -0,0 +1,57 @@
+/*
```
yaozhihang, line 1: Should this class be moved to the `config.project.common` package? Because the CityGML deleter module can also make use of it.  
https://github.com/3dcitydb/importer-exporter/blob/master/impexp-core/src/main/java/org/citydb/core/operation/deleter/controller/Deleter.java#L256

yaozhihang (commented):
@clausnagel Nice. only one remark.

impexp-config/src/main/java/org/citydb/config/project/exporter/ComputeNumberMatched.java
clausnagel: Indeed, good point. Will make a proposal.

clausnagel: addressed in e06d8cd20b0fb047872690a906d69e7717e30dc9 and 6e1787e044610e8317f524bc6f0ecc3357b59414

yaozhihang (approved):

PR #236, opened 2022-01-10
Update H2 database to 2.0.206

clausnagel:
This PR updates the [H2 database](https://github.com/h2database/h2database) to 2.0.206. This is the latest version and a major release of H2. With this release, H2 now works with the latest version of the [JTS](https://github.com/locationtech/jts). This transitive dependency prevented us in the past from updating the PostGIS JDBC driver to its latest version 2021.1.0, because also PostGIS JDBC depends on JTS.

Since H2 is used as local cache for the Importer/Exporter, this PR requires testing to make sure the cache works stable.


clausnagel:
I have removed all database connection parameters that we manually added to increase the performance of the previous H2 version (see `getJDBCUrl` method of the `H2Adapter` class). Most of them are not supported in the latest version anymore, but there might be new or other connection parameters that we should consider in our tests (e.g. `AUTO_ANALYZE=0` might be a candidate to add).

yaozhihang (approved):

Issue #224, opened 2021-11-03
Enhance the `Validate XML query` process


yaozhihang:
The `Validate XML query` process only checks whether the XML query entered into the input field *conforms to the XML schema*. Should we also compile the query to identify more possible errors when clicking this button (like incorrect names and namespaces of features and attributes, incorrect XPath expressions, data type issues, etc.)? Not sure if this is easy to implement though.

clausnagel:
Implemented with #238

PR #238, opened 2022-02-03
Improve XML query validation

clausnagel:
This PR solves #224.

### DIFF
--- a/impexp-client-gui/src/main/java/org/citydb/gui/operation/common/filter/XMLQueryView.java
+++ b/impexp-client-gui/src/main/java/org/citydb/gui/operation/common/filter/XMLQueryView.java
@@ -367,6 +367,20 @@ public class XMLQueryView extends FilterView<QueryConfig> {
             log.error("Validation aborted due to fatal errors.");
         }
 
+        if (errors[0] == 0) {
+            if (connectToDatabase(Language.I18N.getString("filter.dialog.xml.validate.connect"))) {
+                try {
+                    buildQuery();
+                } catch (QueryBuildException e) {
+                    errors[0]++;
+                    log.error("Invalid content: " + e.getMessage(), e.getCause());
+                }
+            } else {
+                log.warn("No database connection established. Aborting validation.");
+                return;
+            }
+        }
+
         if (errors[0] > 0) {
             log.warn(errors[0] + " error(s) reported while validating the XML query.");
         } else {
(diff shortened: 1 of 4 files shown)


impexp-client-gui/src/main/java/org/citydb/gui/operation/common/filter/XMLQueryView.java
```diff
@@ -530,5 +534,30 @@ private boolean isDefaultDatabaseSrs(DatabaseSrs srs) {
                 || (databaseController.isConnected()
                 && srs.getSrid() == databaseController.getActiveDatabaseAdapter().getConnectionMetaData().getReferenceSystem().getSrid());
     }
+
+    private boolean connectToDatabase(String message) {
```
yaozhihang, line 538: This method is very useful. How about move it to a util class for public use?

clausnagel: Hm, what about creating a `DatabaseConnectionDialog` in _org.citydb.gui.dialog_ with a static `boolean show(message, viewController)` method. It would return true or false depending on whether the database connection could be established or not (so you would not know whether the connection was not established because the user clicked 'Cancel' or because there were other issues when trying to connect).

yaozhihang: sounds good.

clausnagel: Ok, please check 36f5a5334c9c412e204636e02345562157164495

yaozhihang: Nice

clausnagel:
@yaozhihang, could you please check once more. I had to do a small change because the following query did not throw an error message when being validated:

```xml
  <typeNames>
    <typeName xmlns:bldg="http://www.opengis.net/citygml/building/2.0">bldg:Building</typeName>
  </typeNames>
  <filter>
    <propertyIsLike wildCard="*" singleCharacter="." escapeCharacter="\" matchCase="true">
      <valueReference>core:creationDate</valueReference>
      <literal>1998</literal>
    </propertyIsLike>
  </filter>
</query>
```

yaozhihang:
Yes, this change is good and makes sense. It's also very hard to catch every invalid content. For example, the following query is also invalid due to the incorrect SQL syntax, and no errors are thrown by the the validator. 
```xml
  <typeNames>
    <typeName xmlns:bldg="http://www.opengis.net/citygml/building/2.0">bldg:Building</typeName>
  </typeNames>
  <filter>
    <sql>
      <select>[invalid sql]</select>
    </sql>
  </filter>
</query>
```

clausnagel:
Yes, I agree. At least we can catch much more issues than before :-) Are you happy to merge?

yaozhihang:
yes

yaozhihang (approved):

Issue #104, opened 2019-10-08
recalculate feature envelope while exporting CityGML with LoD-Filter


yaozhihang:
While running a CityGML-Export, the envelope information of each feature are currently read from the `ENVELOPE` column of the `CITYOBJECT` table and dann propagated to the Cityobject's `boundedBy` property. This may result in an inconsistency problem when the LoD-filter is activated. 

For example, a CityGML-Building dataset could have LOD0 and LOD2 representations in the database at the same time. When exporting Buildings with only LOD0 representation, the envelope of each exported CityGML feature should not be retrived from the CITYOBJECT table directly, but shall be recalculated based on the LOD0 geometries of the respective building object.

clausnagel:
Implemented with #239

PR #239, opened 2022-02-04
Recalculate bbox when LoD filter is used for CityGML exports

clausnagel:
This PR addresses #104.

### DIFF
--- a/impexp-core/src/main/java/org/citydb/core/operation/exporter/concurrent/DBExportWorker.java
+++ b/impexp-core/src/main/java/org/citydb/core/operation/exporter/concurrent/DBExportWorker.java
@@ -56,6 +56,7 @@ import org.citydb.util.event.global.*;
 import org.citygml4j.builder.jaxb.CityGMLBuilder;
 import org.citygml4j.model.gml.base.AbstractGML;
 import org.citygml4j.model.gml.feature.AbstractFeature;
+import org.citygml4j.util.bbox.BoundingBoxOptions;
 
 import java.sql.Connection;
 import java.sql.SQLException;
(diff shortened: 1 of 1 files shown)


impexp-core/src/main/java/org/citydb/core/operation/exporter/concurrent/DBExportWorker.java
```diff
@@ -184,6 +190,11 @@ private void doWork(DBSplittingResult work) {
 					AbstractGML object = exporter.exportObject(work.getId(), work.getObjectType());
 					if (object instanceof AbstractFeature) {
 						feature = (AbstractFeature) object;
+
+						if (useLodFilter && feature.isSetBoundedBy()) {
+							feature.setBoundedBy(feature.calcBoundedBy(bboxOptions));
```
yaozhihang, line 195: The feature here is top-level. How about sub-features, should their bbox also be recalculated?

clausnagel: e8097757143c409f6f67d3a2da7650d71fd52644

yaozhihang:
There is an option `generalOptions.getEnvelope().isUseEnvelopeOnCityModel()`. Maybe it should also be taken into account.

clausnagel:
> There is an option `generalOptions.getEnvelope().isUseEnvelopeOnCityModel()`. Maybe it should also be taken into account.

Well, this option controls whether a `gml:Envelope` shall be created for the root `<CityModel>` in the export dataset. Since `<CityModel>` this is the first line written to the dataset, we cannot compute its envelope based on the exported city objects but only on the envelopes stored in the database.

Or did you mean something else?

yaozhihang:
Yes, that is what I mean. OK, it sounds impossible to reccalculate the CityModel envelope based on the LOD filter.

yaozhihang (approved):

clausnagel:
Well, not impossible in general. But this would require changing the dataset after it has been written. This cannot be done if we write to a stream though. In such cases, we would need some temporary export datasets which is only sent to the stream once it has been adapted. Thus, the stream would have to wait long for a response... So no easy and nice solution, at least.

yaozhihang:
yes, we could improve this in future in a separate PR once we find a better solution.

PR #243, opened 2022-03-03
Avoid duplicate implicit geometries

clausnagel:
This PR proposes to change the way `gml:relativeGeometry` properties of implicit geometries are imported so that duplicate entries are avoided for the `implicit_geometry` table. 

The current importer implementation makes sure that a `gml:relativeGeometry` is only imported once if it is reused within the _same input file_. If, however, the same `gml:relativeGeometry` is also used in another input file, it will be imported again.

With this PR, the importer first checks whether the `gml:relativeGeometry` already exists in the `implicit_geometry` table. If so, it will not be imported again. Like for the current implementation, this check is _only_ based on the `gml:id` of the relative geometry. To be able to check the `gml:id` of previously imported relative geometries, a new `gmlid` column has to be added to the `implicit_geometry` table. This new column is proposed in https://github.com/3dcitydb/3dcitydb/pull/79 and is required for this PR to work.

### DIFF
--- a/impexp-core/src/main/java/org/citydb/core/operation/importer/database/content/DBImplicitGeometry.java
+++ b/impexp-core/src/main/java/org/citydb/core/operation/importer/database/content/DBImplicitGeometry.java
@@ -27,57 +27,66 @@
  */
 package org.citydb.core.operation.importer.database.content;
 
-import org.citydb.config.Config;
 import org.citydb.config.geometry.GeometryObject;
 import org.citydb.core.database.schema.SequenceEnum;
 import org.citydb.core.database.schema.TableEnum;
 import org.citydb.core.database.schema.mapping.MappingConstants;
 import org.citydb.core.operation.common.xlink.DBXlinkLibraryObject;
-import org.citydb.core.operation.common.xlink.DBXlinkSurfaceGeometry;
 import org.citydb.core.operation.importer.CityGMLImportException;
 import org.citydb.core.operation.importer.util.ConcurrentLockManager;
 import org.citydb.core.operation.importer.util.ExternalFileChecker;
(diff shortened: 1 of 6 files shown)


MattSmillie (approved):
Can't comment on the code, however, description looks good to me

impexp-core/src/main/java/org/citydb/core/operation/importer/database/content/DBImplicitGeometry.java
```diff
@@ -27,213 +27,276 @@
...
-				rs = null;
+		return new AbstractMap.SimpleEntry<>(implicitGeometryId, isGlobal);
+	}
+
+	private Map.Entry<Long, Boolean> lookupLibraryObject(String libraryObject) throws SQLException {
+		String lookupKey = "#libraryObject#" + libraryObject;
+		long implicitGeometryId = importer.getObjectId(lookupKey);
+
+		if (implicitGeometryId < 0) {
+			PreparedStatement ps = getOrCreateRelativeGeometryLookup();
```
yaozhihang, line 269: shoud use `getOrCreateLibraryObjectLookup` instead?

yaozhihang (commented):
@clausnagel thanks. One small thing (see my comment)

impexp-core/src/main/java/org/citydb/core/operation/importer/database/content/DBImplicitGeometry.java
clausnagel: Of course! Thanks, fixed with 1064757eaacc0e4327ba661c1fc85dfed1f89922

yaozhihang (approved):

PR #244, opened 2022-03-08
Fix initialization of GUI extensions

clausnagel:
This PR fixes the initialization of GUI extensions. When using a `PreferencesExtension`,  the `getPreferences` method was invoked _before_ the `initGuiExtension` method. However, the `initGuiExtension` is meant as starting point for a GUI plugin and, thus, should be _invoked first_. The idea is, for example, that the preferences object to be returned by the `getPreferences` method can be created inside the `initGuiExtension` method.

### DIFF
--- a/impexp-client-gui/src/main/java/org/citydb/gui/GuiCommand.java
+++ b/impexp-client-gui/src/main/java/org/citydb/gui/GuiCommand.java
@@ -127,23 +127,22 @@ public class GuiCommand extends CliCommand implements StartupProgressListener {
     }
 
     private void initializeViewComponents(ImpExpGui impExpGui, Config config) {
-        // create database plugin
-        DatabasePlugin databasePlugin = new DatabasePlugin(impExpGui, config);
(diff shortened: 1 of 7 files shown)


yaozhihang (commented):
Hm, this MR changes the API and not work with the current VCS delete plugin anymore.  I think the `view` and `preferences` should be initialized in the constructor function. This would also make sure that the methods like `loadSettings` and `switchLocale` will not throw NPE, if the `initGuiExtension` has not been invoked beforehand.

clausnagel:
Thanks @yaozhihang. The API is not changed by this PR but instead correctly implemented. A plugin may still choose to use a constructor to initalize the `view` and `preferences` objects. The issue is that `getPreferences` is currently invoked _before_ `initGuiExtension`  - and this violates the API. The VCS delete plugin easily works with this PR when creating the `preferences` object in the `initGuiExtension` method or the constructor.

yaozhihang:
you are right. But we need a workaround allowing the current Delete Plugin to still work with this PR? Otherwise, we would need a new major version for the Importer/Exporter?

PR #249, opened 2022-04-11
Replace all IDs with UUIDs on export

clausnagel:
This PR adds an option to replace all identifiers of features and geometries with UUIDs when exporting to CityGML/CityJSON. A possible use case is when you want to share your data but at the same time hide your internal object IDs.

I added a new "Object identifier" entry to the export preferences. However, the settings could also be merged into "General" instead. I thought this way it's more consistent with the import preferences where we also have an "Object identifier" entry. Can be discussed.

### DIFF
--- a/impexp-config/src/main/resources/org/citydb/config/i18n/language_en.properties
+++ b/impexp-config/src/main/resources/org/citydb/config/i18n/language_en.properties
@@ -531,7 +531,7 @@ pref.export.general.label.cityModel=Calculate and export bounding box for entire
 pref.export.general.label.useTileExtent=Use tile extent when tiling is enabled
 
 pref.export.id.label.replace=Replace all identifiers by UUIDs
-pref.import.id.label.prefix=Prefix
+pref.export.id.label.prefix=Prefix
 
 pref.export.group.label.exportMember=Export group members as references only
 pref.export.group.label.exportMember.description=<html>Filter settings are <b>not</b> applied</html>
--- a/impexp-config/src/main/resources/org/citydb/config/i18n/language_en.properties
+++ b/impexp-config/src/main/resources/org/citydb/config/i18n/language_en.properties
@@ -404,6 +404,7 @@ pref.tree.import.log=Import log
 pref.tree.import.xmlValidation=XML validation
 pref.tree.import.resources=Resources
 pref.tree.export.general=General
+pref.tree.export.id=Object identifier
 pref.tree.export.group=CityObjectGroup
 pref.tree.export.address=Address
 pref.tree.export.appearance=Appearance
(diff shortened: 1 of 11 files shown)


impexp-core/src/main/java/org/citydb/core/operation/exporter/database/content/CityGMLExportManager.java
```diff
@@ -230,6 +235,10 @@ private AbstractGML processObject(AbstractGML object) throws CityGMLExportExcept
 				// trigger export of textures if required
 				if (isLazyTextureExport() && config.getExportConfig().getAppearances().isSetExportAppearance())
 					getExporter(DBLocalAppearance.class).triggerLazyTextureExport(feature);
+
+				if (idReplacer != null) {
+					feature.accept(idReplacer);
```
yaozhihang, line 240: This should not take place before putting the feature ids into the cache.  It'll cause empty cityobjectmembers when exporting CityObjectGroup

impexp-config/src/main/resources/org/citydb/config/i18n/language_en.properties
```diff
@@ -529,6 +530,9 @@ pref.export.general.feature.none=Do not export object bounding boxes
 pref.export.general.label.cityModel=Calculate and export bounding box for entire dataset
 pref.export.general.label.useTileExtent=Use tile extent when tiling is enabled
 
+pref.export.id.label.replace=Replace all identifiers by UUIDs
+pref.import.id.label.prefix=Prefix
```
yaozhihang, line 534: import -> export

yaozhihang (commented):
Good feature. see my comments

impexp-config/src/main/resources/org/citydb/config/i18n/language_en.properties
clausnagel: Fixed with 3d1178c5379888a5ff1193ccbb4eaac62fe44a66

impexp-core/src/main/java/org/citydb/core/operation/exporter/database/content/CityGMLExportManager.java
clausnagel: Nice catch. The WFS, however, requires the replaced IDs in the cache (e.g., see [here](https://github.com/3dcitydb/web-feature-service/blob/fb4b38f238d2c2f3f8cb3dd4dce1e20f778589bc/src/main/java/vcs/citydb/wfs/operation/getfeature/citygml/CityGMLWriter.java#L314)). So, to fix this, both the original and the replaced IDs are now cached. See 5fca3c003621532650f3195f7e2baa2ebbaa8df6

PR #250, opened 2022-04-15
Enhance tiling output options

clausnagel:
This PR extends the options for defining the subdirectories and filenames for storing tiles when tiling is enabled for CityGML exports.

In contrast to the fixed and limited choices of the current implementation, a user can now freely combine pre-defined tokens such as `%ROW%` and `%COLUMN%` with string literals to define the subdirectories and filenames. Even nested subdirectories are supported (e.g., "`%ROW%`/`%COLUMN%`"). In addition, a format string can be defined for each token, for example, to cut the decimal places from the coordinates of the tile extent. The notation for format strings supports and must comply with the Java notation for [String.format()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Formatter.html).

image


yaozhihang (approved):
Very nice enhancement.

PR #253, opened 2022-05-05
Docker java update

BWibo:
* Updates the Docker images to Java 17
* Gradle wrapper was updated to 7.2 -> 7.4.2 for compatibility with Java 17, [see this](https://docs.gradle.org/current/userguide/compatibility.html).
* Switched to [`bellsoft/liberica-openjdk-alpine`](https://bell-sw.com/pages/downloads/) for Alpine images to avoid error during `vis-export` with several other Alpine JDK base images (openjdk, amazon correto, ...), see below:

   ```text
   [16:37:57 ERROR] Caused by: java.io.IOException: error=2, No such file or directory
   [16:37:57 ERROR] COLLADA2GLTF failed to convert '/data/temp/glTF/Tiles/0/0/209/GMLID_BUI180021_673_14143.dae'.
   [16:37:57 ERROR] Caused by: java.io.IOException: Cannot run program "/opt/impexp/contribs/collada2gltf/COLLADA2GLTF-v2.1.4-linux/COLLADA2GLTF-bin" (in directory "/data/temp/glTF/Tiles/0/0/209"): error=2, No such file or directory
   [16:37:57 ERROR] Caused by: java.io.IOException: error=2, No such file or directory
   [16:37:58 ERROR] COLLADA2GLTF failed to convert '/data/temp/glTF/Tiles/0/0/228/GMLID_21614432_375009_394.dae'.
   [16:37:58 ERROR] Caused by: java.io.IOException: Cannot run program "/opt/impexp/contribs/collada2gltf/COLLADA2GLT-v2.1.4-linux/COLLADA2GLTF-bin" (in directory "/data/temp/glTF/Tiles/0/0/228"): error=2, No such file or directory
   ```

### DIFF
--- a/properties.gradle
+++ b/properties.gradle
@@ -28,5 +28,5 @@ ext {
     appCliName = 'impexp'
 
     // documentation
-    docUrl = 'https://3dcitydb-docs.readthedocs.io/en/version-2022.1/'
+    docUrl = 'https://3dcitydb-docs.readthedocs.io/en/latest/'
 }
--- a/properties.gradle
+++ b/properties.gradle
@@ -28,5 +28,5 @@ ext {
     appCliName = 'impexp'
 
     // documentation
-    docUrl = 'https://3dcitydb-docs.readthedocs.io/en/latest/'
+    docUrl = 'https://3dcitydb-docs.readthedocs.io/en/version-2022.1/'
 }
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,8 @@
 
 # Fetch & build stage #########################################################
 # ARGS
-ARG BUILDER_IMAGE_TAG='11.0.12-jdk-slim'
-ARG RUNTIME_IMAGE_TAG='11.0.12-jre-slim'
+ARG BUILDER_IMAGE_TAG='17-jdk-slim'
+ARG RUNTIME_IMAGE_TAG='17-slim'
 
 # Base image
 FROM openjdk:${BUILDER_IMAGE_TAG} AS builder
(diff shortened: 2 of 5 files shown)


properties.gradle
```diff
@@ -28,5 +28,5 @@ ext {
     appCliName = 'impexp'
 
     // documentation
-    docUrl = 'https://3dcitydb-docs.readthedocs.io/en/version-2022.1/'
```
yaozhihang, line 31: If we want to merge this to release 5.2 branch, we should use the release version of the doc `version-2022.1`

yaozhihang (commented):
@BWibo thanks, very interesting issue. Have you investigated why other Alpine JDK base images (openjdk, amazon correto, ...) cause the error?

properties.gradle
BWibo: `version-2022.2`, not `version-2022.1` right?

BWibo:
@yaozhihang Yes, I discussed it with @clausnagel via eMail. I assume that there are some libraries missing in the Alpine base image. I didn't look into it closely, but I have had this issue with other Alpine based images too.

properties.gradle
BWibo: 👍

yaozhihang (approved):
Now the PR looks good to me. 

I have actually one concern about the future versions of the `liberica-openjdk-alpine` image, which could also produce the same error!? Maybe we should think about if we drop the support of Alpine image based version of the importer/exporter or all 3dcitydb software.

BWibo:
Hey, that is possible, but at least for the LTS 17 image rather unlikely. I think as long as we stick to Java 17 we're good.
In general it would be good to keep the Alpine image, as it is smaller than the Debian version and Alpine images are well accepted in the community. 
If we see that this is causing lot's of maintenance in the future, we can still drop it.

clausnagel (approved):

PR #258, opened 2022-06-10
Prevent XXE vulnerabilities

clausnagel:
This PR adds necessary code to prevent [XXE vulnerabilities](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html) when parsing XML files.

XML files are not just CityGML input files. But also the config file of the Importer/Exporter, XML query expressions, XSD schema files, and XSLT stylesheets are XML-based. So, various classes are affected by this PR.

- Note 1: The `ConfigUtil` had to be moved from impexp-config to impexp-util to be able to use the new `SecureXMLProcessors` class.
- Note 2: This PR uses the 2.12.1-SNAPSHOT version of citygml4j, which also contains XXE fixes (see [here](https://github.com/citygml4j/citygml4j/commit/246e34a372ed4a8c4082de7170153b3014d55961)).

### DIFF
--- a/impexp-util/src/main/java/org/citydb/util/xml/SecureXMLProcessors.java
+++ b/impexp-util/src/main/java/org/citydb/util/xml/SecureXMLProcessors.java
@@ -50,6 +50,8 @@ public class SecureXMLProcessors {
         factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
         factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
         factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
+        factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
+
         return factory;
     }
 
--- a/impexp-util/src/main/java/org/citydb/util/xml/SecureXMLProcessors.java
+++ b/impexp-util/src/main/java/org/citydb/util/xml/SecureXMLProcessors.java
@@ -1,0 +1,75 @@
+/*
+ * 3D City Database - The Open Source CityGML Database
+ * https://www.3dcitydb.org/
+ *
+ * Copyright 2013 - 2022
+ * Chair of Geoinformatics
+ * Technical University of Munich, Germany
+ * https://www.lrg.tum.de/gis/
+ *
+ * The 3D City Database is jointly developed with the following
+ * cooperation partners:
+ *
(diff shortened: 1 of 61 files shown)


impexp-util/src/main/java/org/citydb/util/xml/SecureXMLProcessors.java
```diff
@@ -0,0 +1,75 @@
...
+    private SecureXMLProcessors() {
+    }
+
+    public static SAXParserFactory newSAXParserFactory() throws SAXNotSupportedException, SAXNotRecognizedException, ParserConfigurationException {
+        SAXParserFactory factory = SAXParserFactory.newInstance();
+        factory.setXIncludeAware(false);
+        factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+        factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
+        factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
```
yaozhihang, line 52: Is the following code also needed? see [here](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#jaxb-unmarshaller)
``` java
factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
```

yaozhihang (approved):
Thanks @clausnagel. Very important fix. I have only one remark from my review.

impexp-util/src/main/java/org/citydb/util/xml/SecureXMLProcessors.java
clausnagel: Not sure whether it's actually _needed_. I've found different examples on the web, and some even just use `factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);`. 

I am not against adding it. It's just that not all properties are supported by all XML implementations in every use case. So, could you please check that we do not get `property-is-not-supported` exceptions when adding it (e.g. when creating a validating or a non-validating reader)?

yaozhihang: Done with a82199a

yaozhihang (approved):

Issue #257, opened 2022-06-01
Export timeout


BWibo:
I have a big 3DCityDB (v4.3.0) instance containing more than 1.5 million buildings. 
When I try to export (ImpExp v5.2.0,  both GUI and CLI via Docker) a bigger area (City of Munich) delimited by a polygon from that database using an XML query, I am facing following issue:

When I try to use the XML query listed below including `<tiling>` the export terminates after some time with this error, which probably results of a connection being closed after being idle for too long.

```
Jun 01, 2022 12:50:23 PM org.apache.tomcat.jdbc.pool.ConnectionPool terminateTransaction
WARNING: Unable to terminate transaction, connection will be closed.
org.postgresql.util.PSQLException: This connection has been closed.
        at org.postgresql.jdbc.PgConnection.checkClosed(PgConnection.java:883)
        at org.postgresql.jdbc.PgConnection.getAutoCommit(PgConnection.java:843)
        at org.apache.tomcat.jdbc.pool.ConnectionPool.terminateTransaction(ConnectionPool.java:916)
        at org.apache.tomcat.jdbc.pool.ConnectionPool.shouldClose(ConnectionPool.java:954)
        at org.apache.tomcat.jdbc.pool.ConnectionPool.returnConnection(ConnectionPool.java:1018)
        at org.apache.tomcat.jdbc.pool.ProxyConnection.invoke(ProxyConnection.java:103)
        at org.apache.tomcat.jdbc.pool.JdbcInterceptor.invoke(JdbcInterceptor.java:109)
        at org.apache.tomcat.jdbc.pool.interceptor.AbstractCreateStatementInterceptor.invoke(AbstractCreateStatementInterceptor.java:69)
        at org.apache.tomcat.jdbc.pool.JdbcInterceptor.invoke(JdbcInterceptor.java:109)
        at org.apache.tomcat.jdbc.pool.TrapException.invoke(TrapException.java:39)
        at org.apache.tomcat.jdbc.pool.JdbcInterceptor.invoke(JdbcInterceptor.java:109)
        at org.apache.tomcat.jdbc.pool.interceptor.AbstractCreateStatementInterceptor.invoke(AbstractCreateStatementInterceptor.java:69)
        at org.apache.tomcat.jdbc.pool.JdbcInterceptor.invoke(JdbcInterceptor.java:109)
        at org.apache.tomcat.jdbc.pool.DisposableConnectionFacade.invoke(DisposableConnectionFacade.java:81)
[...]

PR #261, opened 2022-07-21
Hotfix export timeout issue

yaozhihang:
Fix the issue #257. 

It seems that the issue is caused by the very long `where` clause in the root select statement. So, putting the long `where` clause in a subquery solves the issue.

### DIFF
--- a/impexp-core/src/main/java/org/citydb/core/operation/exporter/database/content/DBSplitter.java
+++ b/impexp-core/src/main/java/org/citydb/core/operation/exporter/database/content/DBSplitter.java
@@ -238,16 +238,17 @@ public class DBSplitter {
 			hits = getNumberMatched(query, connection);
 		}
 
+		// avoid long "where" clause in the root select statement
+		Table table = new Table(select);
+		select = new Select().addProjection(table.getColumn(MappingConstants.ID))
+				.addProjection(table.getColumn(MappingConstants.OBJECTCLASS_ID))
+				.addProjection(table.getColumn(MappingConstants.GMLID))
+				.addProjection(table.getColumn(MappingConstants.ENVELOPE));
+
 		// add spatial extent
 		if (calculateExtent) {
-			Table table = new Table(select);
-			select = new Select().addProjection(table.getColumn(MappingConstants.ID))
-					.addProjection(table.getColumn(MappingConstants.OBJECTCLASS_ID))
-					.addProjection(table.getColumn(MappingConstants.GMLID))
-					.addProjection(new Function(databaseAdapter.getSQLAdapter().resolveDatabaseOperationName("geom_extent") +
+			select.addProjection(new Function(databaseAdapter.getSQLAdapter().resolveDatabaseOperationName("geom_extent") +
 					"(" + table.getColumn(MappingConstants.ENVELOPE) + ") over", "extent"));
-			if (query.isSetTiling())
-				select.addProjection(table.getColumn(MappingConstants.ENVELOPE));
 		}
 
 		// issue query
--- a/impexp-core/src/main/java/org/citydb/core/operation/exporter/database/content/DBSplitter.java
+++ b/impexp-core/src/main/java/org/citydb/core/operation/exporter/database/content/DBSplitter.java
@@ -514,7 +514,7 @@ public class DBSplitter {
 
 					// send appearance to export workers
 					DBSplittingResult splitter = new DBSplittingResult(appearanceId, appearanceType);
-				//	dbWorkerPool.addWork(splitter);
+					dbWorkerPool.addWork(splitter);
 				} while (rs.next() && shouldRun);
 			}
 		}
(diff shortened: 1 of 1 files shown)


BWibo:
@yaozhihang I just tested this with the workaround described in https://github.com/3dcitydb/importer-exporter/issues/257#issuecomment-1191038324 enabled and disabled and can confirm this works! Thx for the quick fix.

clausnagel:
@yaozhihang, thanks for proposing this hotfix to solve #257. However, I'm not sure whether this is really a "fix". Imho, the query optimizer should deal with "long" `where` clauses automatically, not the application code. Rearranging the query in the application code as proposed might help in the context of #257, but might have no (or even a bad) impact in other scenarios. For instance, the `getNumberMatched` method in `DBSplitter` still uses the unchanged query and yet performs good.

And does your change affect exports from Oracle?

@BWibo, have you already tried and "vacuumed" your database? Maybe this already helps?

impexp-core/src/main/java/org/citydb/core/operation/exporter/database/content/DBSplitter.java
```diff
@@ -238,16 +238,17 @@ private void queryCityObject(FeatureType cityObjectGroupType, Map<Long, DBSplitt
 			hits = getNumberMatched(query, connection);
 		}
 
+		// avoid long "where" clause in the root select statement
+		Table table = new Table(select);
+		select = new Select().addProjection(table.getColumn(MappingConstants.ID))
+				.addProjection(table.getColumn(MappingConstants.OBJECTCLASS_ID))
+				.addProjection(table.getColumn(MappingConstants.GMLID))
+				.addProjection(table.getColumn(MappingConstants.ENVELOPE));
```
clausnagel, line 246: The `ENVELOPE` column may only be added if both `calculateExtent` and `query.isSetTiling()` are true (see previous code). Otherwise, the query will throw errors, for example, when tiling is not enabled.

clausnagel (changes requested):

impexp-core/src/main/java/org/citydb/core/operation/exporter/database/content/DBSplitter.java
BWibo: @clausnagel 

> @BWibo, have you already tried and "vacuumed" your database? Maybe this already helps?

I just did a `vacuum analyze` on the database and tested again with ImpExp `v5.2.0`. 
The export did not get stuck this time! I thought I had done this already, but apparently I forgot to run the vaccum.

If this resolves the issue with the original query, we should at least add a hint to vaccum the database, after importing large datasets.

yaozhihang: oh, puh, the query has problem with Oracle. Tested with both 5.1 and 5.2 and Oracle 12 and 19

```
[11:31:05 ERROR] Failed to query the database.
[11:31:05 ERROR] Caused by: java.sql.SQLException: ORA-22901: VARRAY oder LOB-Attribute eines Objekttyps können nicht verglichen werden
[11:31:05 ERROR] Caused by: oracle.jdbc.OracleDatabaseException: ORA-22901: VARRAY oder LOB-Attribute eines Objekttyps können nicht verglichen werden
[11:31:05 WARN] Database export aborted.
```

clausnagel:
@BWibo, thanks for vacuuming the database. Unfortunately, I again ran into the issue in my export test. So, it seems that vacuuming might help but does not solve the issue.

clausnagel:
When adding `DatabaseConnectionPool.getInstance().purge();` as line 591 to the class `org.citydb.core.operation.exporter.controller.Exporter` (so, as last line of the `finally` block), the export also completes without issues in my tests.

This seems to support that re-arranging the query is not necessary. As far as I understand, `purge()` closes all idle connections. This is of course not what we want in the first place, because we use a pool so that we don't have to physically open connections all the time. But still it shows that there seems to be an issue with pool maintenance and handling of idle connections (maybe even a Tomcat JDBC pool bug?).

impexp-core/src/main/java/org/citydb/core/operation/exporter/database/content/DBSplitter.java
clausnagel: Should we open a separate issue to deal with this error?

yaozhihang: yes, see #262

yaozhihang:
clausnagel's solution looks better. It ensures that the pool produces clean connections for each export.

clausnagel:
It seems that PR #264 also solves the issue #257 and thus makes this PR obsolete. @yaozhihang and @BWibo, can you please test and confirm?

clausnagel:
Discarding this PR due to #263

PR #263, opened 2022-07-26
Alternative hotfix export timeout issue

clausnagel:
This PR proposes an alterantive to #261 for fixing #257.

This hotfix purges the connection pool after every CityGML export. This prevents issues due to defect connections as the one reported in #257. A drawback of this approach is that the number of physical connections created by the pool increases especially when exporting in tiles, which undermines the idea of using a pool.

As identified by @yaozhihang in #261, it seems that the reason for #257 is the long `where` clause of the SQL query due to a polygon with a huge number of vertices encoded as WKT. Somehow a connection cannot be reused after sending this query to the database. I couldn't figure out how to identify such a defect connection and only remove it from the pool. This would be a better solution.

### DIFF
--- a/impexp-core/src/main/java/org/citydb/core/database/connection/DatabaseConnectionPool.java
+++ b/impexp-core/src/main/java/org/citydb/core/database/connection/DatabaseConnectionPool.java
@@ -29,6 +29,7 @@ package org.citydb.core.database.connection;
 
 import org.apache.tomcat.jdbc.pool.DataSource;
 import org.apache.tomcat.jdbc.pool.PoolProperties;
+import org.apache.tomcat.jdbc.pool.PooledConnection;
 import org.citydb.config.Config;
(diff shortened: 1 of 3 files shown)


clausnagel:
It seems that PR #264 also solves the issue #257 and thus makes this PR obsolete. @yaozhihang and @BWibo, can you please test and confirm?

clausnagel:
I slightly changed the implementation. Now, _only_ the connection opened by `DBSplitter` is physically closed and removed from the pool instead of purging _all_ connections.

yaozhihang (approved):

BWibo:
I made a fresh build from #263 and tested again:
* The now vacuumed DB you have access to works fine for me.
* Another instance I setup that has not been vacuumed works fine too.

BWibo (approved):

Issue #225, opened 2021-11-03
Allow to choose multiple CSV files for the Import list filter


yaozhihang:

clausnagel:
I checked this feature wish and it requires quite some changes to the `IdList` core API on which the import list filter is built. Thus, these changes will also have an impact on all plugins using an `IdList`.

Some additional questions:
- Currently, the CSV file to be parsed is shown using a simple `JTextField` in the UI. I guess this should be changed to a `JList` to show multiple files? We could keep a simple `JTextField` but then would need a unqiue separator (for example embrace all files names by double quotes and use a white space inbetween: "fileA" "fileB" "fileC").
- All files must, of course, share the same CSV settings. When doing a preview, I would still just preview the first file, not all files. Agreed?

yaozhihang:
Thanks for looking into this feature wish.
- `JList` looks better. 
- Fully agree

clausnagel:
Solved with #271

PR #271, opened 2022-09-14
Allow ID lists to be composed of one or more CSV files

clausnagel:
This PR allows ID lists to be composed of one or more CSV files. The settings for parsing the CSV content are specified only once and, thus, must be met by all CSV files composing the ID list. The changes are implemented for the import list (GUI + CLI) and the delete list (CLI only).

In GUI mode, the user can select for which of the CSV files a preview should be generated when clicking the `Preview` button. Multi-selection is supported. The default is to generate the preview only for the first CSV file.

In CLI mode, the preview is always just generated for the _first_ CSV files. This behaviour could be changed so that the preview is generated for _all_ files if required. 

Solves #225.

### DIFF
--- a/impexp-client-gui/src/main/java/org/citydb/gui/components/FileListTransferHandler.java
+++ b/impexp-client-gui/src/main/java/org/citydb/gui/components/FileListTransferHandler.java
@@ -160,8 +160,8 @@ public class FileListTransferHandler extends TransferHandler implements DropTarg
     @SuppressWarnings("unchecked")
     @Override
     public void drop(DropTargetDropEvent event) {
-        for (DataFlavor dataFlover : event.getCurrentDataFlavors()) {
-            if (dataFlover.isFlavorJavaFileListType()) {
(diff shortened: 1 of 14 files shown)


impexp-client-gui/src/main/java/org/citydb/gui/components/FileListTransferHandler.java
```diff
@@ -0,0 +1,230 @@
...
+
+    @Override
+    public void dragEnter(DropTargetDragEvent event) {
+        event.acceptDrag(DnDConstants.ACTION_COPY_OR_MOVE);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public void drop(DropTargetDropEvent event) {
+        for (DataFlavor dataFlover : event.getCurrentDataFlavors()) {
```
yaozhihang, line 163: dataFlover -> dataFlavor

yaozhihang (approved):
Nice enhancement. @clausnagel.

PR #274, opened 2022-11-02
Added support for loading plugins and ade-extensions over symlinks

yaozhihang:


impexp-client-cli/src/main/java/org/citydb/cli/ImpExpCli.java
```diff
@@ -535,6 +535,9 @@ private void createPidFile() throws ImpExpException {
 
     private void loadClasses(Path path, Util.URLClassLoader classLoader) throws IOException {
         if (Files.exists(path)) {
+            if (Files.isSymbolicLink(path)) {
+                path = Files.readSymbolicLink(path);
+            }
             try (Stream<Path> stream = Files.walk(path)
```
clausnagel, line 541: Why not just use `Files.walk(path, FileVisitOption.FOLLOW_LINKS)`?

yaozhihang: Yes, it is nicer

clausnagel (approved):

Issue #285, opened 2023-05-31
NoClassDefFoundError: org/citydb/gui/plugin/view/ViewExtension when running the Importer/Exporter CLI


spechtx:
Describe the bug:
When attempting to use the Importer/Exporter command line interface (version 5.3.0), a java.lang.NoClassDefFoundError is thrown with the message org/citydb/gui/plugin/view/ViewExtension.

To Reproduce:

Run the following command:

`java -jar lib/impexp-client-cli-5.3.0.jar import --duplicate-log=/mnt/data/temp/3dcitydb/duplicate.log --import-log=/mnt/data/temp/import.log --input-encoding=utf-8 --log-file=/mnt/data/temp/3dcitydb/logfile.log --log-level=info --import-mode=import_all --worker-threads=16 --db-type=postgresql --db-host=localhost --db-port=5434 --db-name=3DCITYDB --db-schema=citydb --db-username=***** --db-password=***** /mnt/data/temp/3dcitydb/hh/`

See error:
```
[12:22:15 INFO] Starting 3D City Database Importer/Exporter, version 5.3.0
[12:22:15 INFO] Loading plugins
Exception in thread "main" java.lang.NoClassDefFoundError: org/citydb/gui/plugin/view/ViewExtension
```

Expected behavior:
The Importer/Exporter should start and load the plugins without any exceptions.

Environment (please complete the following information):

OS: Ubuntu Server 22.04
Java version: OpenJDK 11
Database: PostgreSQL 15]

spechtx:
I was able to start the import when using clinent-gui instead of client-cli:
`java -jar lib/impexp-client-gui-5.3.0.jar`

spechtx:
Workaround:
Interestingly, when running the command with the client GUI, instead the CLI, the import proceeds without any issues:
`java -jar lib/impexp-client-gui-5.3.0.jar import --duplicate-log=/mnt/data/temp/3dcitydb/duplicate.log --import-log=/mnt/data/temp/import.log --input-encoding=utf-8 --log-file=/mnt/data/temp/3dcitydb/logfile.log --log-level=info --import-mode=import_all --worker-threads=16 --db-type=postgresql --db-host=localhost --db-port=5434 --db-name=3DCITYDB --db-schema=citydb --db-username=***** --db-password=***** /mnt/data/temp/3dcitydb/hh/`

This is unexpected, as I am working in a shell environment and should not require the GUI version to run the import.

clausnagel:
[...]

PR #287, opened 2023-07-07
log the plugin loading NoClassDefFoundError

yaozhihang:
Fix the issue
https://github.com/3dcitydb/importer-exporter/issues/285

### DIFF
--- a/impexp-core/src/main/java/org/citydb/core/plugin/PluginManager.java
+++ b/impexp-core/src/main/java/org/citydb/core/plugin/PluginManager.java
@@ -65,7 +65,7 @@ public class PluginManager {
             for (Plugin plugin : pluginLoader) {
                 plugins.add(plugin);
             }
-        } catch (NoClassDefFoundError e) {
+        } catch (Throwable e) {
             log.error("Failed to load plugins.", e);
             return;
         }
--- a/impexp-core/src/main/java/org/citydb/core/plugin/PluginManager.java
+++ b/impexp-core/src/main/java/org/citydb/core/plugin/PluginManager.java
@@ -42,6 +42,7 @@ import java.util.*;
 
 public class PluginManager {
     private static PluginManager instance;
+    private final Logger log = Logger.getInstance();
     private final List<InternalPlugin> internalPlugins = new ArrayList<>();
     private final List<Plugin> externalPlugins = new ArrayList<>();
     private final List<CliCommand> commands = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


clausnagel:
Thanks, @yaozhihang. Couldn't we alternatively just log exceptions for those plugins that cannot be loaded (via something like `addException`) and load those plugins that can be loaded?

impexp-core/src/main/java/org/citydb/core/plugin/PluginManager.java
```diff
@@ -58,7 +59,18 @@ public static synchronized PluginManager getInstance() {
...
         JAXBContext context = createJAXBContext();
         ServiceLoader<Plugin> pluginLoader = ServiceLoader.load(Plugin.class, loader);
-        for (Plugin plugin : pluginLoader) {
+
+        List<Plugin> plugins = new ArrayList<>();
+        try {
+            for (Plugin plugin : pluginLoader) {
+                plugins.add(plugin);
+            }
+        } catch (NoClassDefFoundError e) {
```
clausnagel, line 68: I suggest catching `Throwable` here

yaozhihang: done with [952a7fd](https://github.com/3dcitydb/importer-exporter/pull/287/commits/952a7fdbffa340c1c159b9901d10a09f37eed67e)

yaozhihang:
> 

Hm, i think not possible. The exception is thrown in the method `hasNextService` of the ServiceLoader and no plugin info is returned. You have ideas?

clausnagel:
> > 
> 
> Hm, i think not possible. The exception is thrown in the method `hasNextService` of the ServiceLoader and no plugin info is returned. You have ideas?

Yes, I also tested the iterator interface of the ServiceLoader and unfortunately the error is already thrown in the `hasNext` instead of the `next` method. So, it's not really possible to skip just those plugins with errors and load the others.

I nevertheless propose a small change. I think it would be better to use the exception handling of the `PluginManager` class. This way, the CLI fails on errors (default behaviour), while the error is shown in the console window when using the GUI.

[proposal.patch](https://github.com/3dcitydb/importer-exporter/files/11998989/proposal.patch)

yaozhihang:
> > > 
> > 
> > 
> > Hm, i think not possible. The exception is thrown in the method `hasNextService` of the ServiceLoader and no plugin info is returned. You have ideas?
> 
> I also tested the iterator interface of the ServiceLoader and unfortunately the error is already thrown in the `hasNext` instead of the `next` method. So I can confirm that it's not really possible to skip just those plugins with errors and load the others.
> 
> I nevertheless propose a small change. I think it would be better to use the exception handling of the `PluginManager` class. This way, the CLI fails on errors (default behaviour), while the error is shown in the console window when using the GUI.
> 
> [proposal.patch](https://github.com/3dcitydb/importer-exporter/files/11998989/proposal.patch)

thanks @clausnagel. The Proposal is in this commit [d7ff2c3](https://github.com/3dcitydb/importer-exporter/pull/287/commits/d7ff2c3f74bdc66e2dd748ea75540c872e35b170)

clausnagel (approved):

PR #313, opened 2024-11-28
refactor: removed complex conditional statement in toQueryConfig function in QueryOption class

MeetMaratha:
In this PR I have extracted a complex conditional to a separate function and gave it proper name for removing complex conditional smell in the code base.


clausnagel:
I don't see the benefit of introducing this change.

PR #314, opened 2024-11-28
refactor: removed complex method smell by reducing cyclomatic complexity of toQueryConfig function in QueryOption class

MeetMaratha:
In this PR I have extracted a complex method to a separate set of functions and gave it proper name for reducing the cyclomatic complexity of the original function in the code base.


clausnagel:
I don't see the benefit of introducing this change.

PR #315, opened 2024-11-28
refactor: removed feature envy smell from Tile class and pushed the method to BoundingBox class

MeetMaratha:
In this PR I have moved the computation of if a point on a tile is inside the bounding box to the BoundingBox class from the Tile class. I also added some intermediate variables to make things more clear for the developers.


clausnagel:
The `isOnTile` check is not the same as a `containsPoint` check for a bounding box. So rejecting this PR.

PR #316, opened 2024-11-28
refactor: removed unclear variable name smell by renaming preview variable to isPreviewOn

MeetMaratha:
In this PR I have renamed a variable to have a more defining name.


clausnagel:
I don't see a need for introducing this change.

PR #317, opened 2024-11-28
Refactor/remove multi faceted abstraction smell

MeetMaratha:
In this PR I tried to solve multi-faceted abstraction smell present in the DirectoryScanner class. This class was performing multiple tasks, so I have extracted them into separate classes with there being an abstraction for processing file such that new ways of processing file can be added easily in future by extending FileProcessor class and adding logic for the new processing function.

I built the program using command gradle installDist, and it built completely without any issues and I was able to run the GUI without any errors.


clausnagel:
There is currently no plan for adding new file formats, so an extensible file processing logic is not needed. What is the use case?

